PR #118, opened 2024-03-28
CRISTAL-165: Initialize the main layout

tkrieck:
# Jira URL
https://jira.xwiki.org/browse/CRISTAL-165

# Changes

## Description

* Start the layout based on prototypes and discussions in forums and design page (Version A for now)

## Clarifications

* Proposal: https://design.xwiki.org/xwiki/bin/view/Proposal/CristalMainlayoutandnavigationUIproposal
* Developed using the config XWikiSL.
* TODOS: all todos for development right now are listed with comments prefixed with "TODO CRISTAL-165:"
* Especially important, the !important rules found in CSS are temporary, they depend on the CSS structure being refactored.

# Screenshots & Video

localhost_9000_XWikiSL_

# Executed Tests

-

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


ds/shoelace/src/vue/x-navigation-drawer.vue
```diff
@@ -6,4 +6,8 @@
   </div>
 </template>
 
-<style scoped></style>
+<style scoped>
+:global(#sidebar) {
+  background-color: var(--sl-color-gray-100);
+}
+</style>
```
manuelleduc, line 13: There are several issues with this approach:
1. using `:global` is always dangerous as it could impact unexpected elements
2. we are making the assumption that this component will be called with the `sidebar` is, and we need to reason in terms of Vue component encapsulation (i.e., preferably zero-knowledge of the parents and children)

Some alternative design (brainstorming, we'll need to define best practices).

## Option 1 - class based

```
<template>
  <div class="background">
    <slot></slot>
  </div>
</template>

<style scoped>
.background {
  background-color: var(--sl-color-gray-100);
}
</style>
```

## Option 2: module based

see https://vuejs.org/api/sfc-css-features.html#custom-inject-name

```
<template>
  <div :class="xnavigationdrawer.background">
    <slot></slot>
  </div>
</template>

<style scoped module="xnavigationdrawer">
.background {
  background-color: var(--sl-color-gray-100);
}
</style>
```

The difference between the two options are subtle but could become important over time.
With option 1, the root div of the element has a `background` class, which could be used by global CSS sheets (i.e., could be used by css extensions)
With option 2, the root div element is decorated with a random class (e.g., `_background_fpq0f_2`), and a corresponding CSS rule is generated (e.g., `._background_fpq0f_2[data-v-074f776b] {background-color: var(--sl-color-gray-100);}`.
In this case, the class is opaque and can't be anticipated (e.g., it will likely change from one version release to another), meaning that we can prevent global CSS sheets to interfere with the element, avoiding introducing "APIs" that are hard to change later.

Note: I know this is rather technical, I'm writing as a support for future discussions.

lib/src/c-index.vue
```diff
@@ -36,57 +36,82 @@ export default defineComponent({
 <template>
   <CTemplate name="view" />
 </template>
+<!-- TODO CRISTAL-165: these styles need to me moved to a generic CSS that loads last in the main template-->
```
manuelleduc, line 39: +1 we'll need to discuss how to organise those rules.

skin/src/vue/c-config-menu.vue
```diff
@@ -1,12 +1,13 @@
...
 import CTemplate from "./c-template.vue";
 import logo from "../images/logo-x.png";
+import xdevCog from "../images/dev-cog.svg";
 </script>
 
 <template>
   <x-dialog width="auto" :logo="logo" title="Configuration">
     <template #activator="{ props }">
-      <x-img :width="50" :src="logo" v-bind="props" />
+      <x-img :src="xdevCog" width="24px" v-bind="props" />
```
manuelleduc, line 10: TODO: discuss how to handle icons.
I can thinl of three options:
1. each design system has its own icon set
1. we can define many icon sets, and they can be used by all design systems
1. we define a single icon set and use it in all design systems

I'm +1 for option 1 or 2 and -1 for option 3.
But, we'll need to make a proposal for that point as it is quite structural.

skin/src/vue/c-content.vue
```diff
@@ -114,14 +114,22 @@ onUpdated(() => {
     </template>
     <template v-else>
       <!-- eslint-disable vue/no-v-html -->
-      <XBreadcrumb></XBreadcrumb>
+      <div class="breadcrumb">
+        <XBreadcrumb></XBreadcrumb>
+      </div>
```
manuelleduc, line 119: Note, another options is:
```suggestion
<XBreadcrumb class="breadcrumb"></XBreadcrumb>
```

And to change the css to :
```css
.breadcrumb {
  display: block;
  padding: 8px;
}
```

manuelleduc: See https://vuejs.org/guide/essentials/class-and-style.html for some explanations regarding how class/style works with components.

skin/src/vue/c-config-menu.vue
tkrieck: Option 3 is probably the more laborious one also. It would fall on to us to provide the icons to the DS. When updating the DS version, for example, we would need to check icons for new components and keep tabs on the current ones if something has changed.

Option 2 seems to be the most flexible. Probably we will need support for three kinds of pack of icons: 
1 - The default one for the DS (not our responsibility, it would be loaded by the DS itself). 
2 - Cristal icons, for its own internal components. These can be from the same icon pack as the DS, but its availability can be unreliable in case of a DS change.
3 - User icons. Loaded by some type of customization on top of the two already loaded.

Depending on the installation, we would have different scenarios of icons being loaded. Best case scenario, all three are the one and the same. Worst case, three different packs being loaded, a lot more resources being consumed and possibly a lot of visual mismatch between them.

Option 1 seems to be the most restrictive, but I might be missing something. I am unsure how Cristal-only components would work with icons class names being defined by the DS. Probably it would need a solution like the IconTheme extension from XS, but that comes with its drawbacks too, like always needing to have fallbacks for each icon.

So for me, 2 is the most interesting, but with a sane default to keep it light on resources and visual integrity.

lib/src/c-index.vue
tkrieck: Indeed, I don't know yet the best way to accomplish this in Vue.

ds/shoelace/src/vue/x-navigation-drawer.vue
tkrieck: I agree with the premise, but unsure of the best way forward between two options. I am more acquainted with method 1. But as this a Vue project, perhaps there's more value with option 2 that I am missing.

skin/src/vue/c-view.vue
```diff
@@ -29,10 +29,31 @@ import CTemplate from "./c-template.vue";
...
+
       <UIX uixname="view.after" />
     </div>
   </x-layout>
 </template>
+<style scoped>
+.collapsed-sidebar {
+  display: none;
+  background: var(--sl-color-gray-100);
+  grid-area: collapsed-sidebar;
```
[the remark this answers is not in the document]
tkrieck: The grid layout was not being updated with simple declaration of` display: none` on the containing element. The screen space was still being visible, but blank.

I could update the layout to not use a grid layout, but I feel grid gives us more flexibility.

ds/shoelace/src/vue/x-navigation-drawer.vue
manuelleduc: Let's go with option 2 for now as it's easier to switch back to option 1 if we hit an issue, than the other way around.

lib/src/c-index.vue
manuelleduc: Steps:
1. add `import "./index.css";` in the imports of the script section
2. introduce a `index.css` file aside this component with the following content:

```css
html,
body,
#app,
main,
#content,
#xwikicontent {
  height: 100vh;
  overflow: hidden;
  position: relative;
}

/* TODO CRISTAL-165: this rule !important should be removed when reorganizing the css as the comment above instructs. */
html {
  overflow: hidden !important;
}

#xwikicontent {
  overflow: auto;
}

.wrapper {
  height: 100vh;
  display: grid;
  grid-template-columns: 250px 1fr;
  grid-template-areas: "sidebar content";
}

.wrapper:has(.right-sidebar) {
  grid-template-columns: 250px 1fr 250px;
  grid-template-areas: "sidebar content right-sidebar";
}

/* TODO CRISTAL-165: move this to shoelace component scoped styles*/
.right-sidebar {
  background-color: var(--sl-color-gray-100);
  padding: 8px;
  overflow: auto;
}

#content {
  display: flex;
  flex-flow: column;
  grid-area: content;
}

#xwikicontent {
  overflow: auto;
  height: 100%;
  padding: 8px;
  box-sizing: border-box;
}

#header {
  grid-area: header;
}

#sidebar {
  grid-area: sidebar;
}

footer {
  grid-area: footer;
  background-color: #005ed8;
}

@media (max-width: 500px) {
  /* TODO CRISTAL-165: all of these !important declarations should be unecessary when the CSS structure is refactored */
  .wrapper {
    grid-template-columns: 24px 1fr !important;
    grid-template-areas: "collapsed-sidebar content" !important;
  }

  .collapsed-sidebar {
    display: block !important;
  }

  #sidebar {
    display: none !important;
  }

  .right-sidebar {
    display: none;
  }
}
`` `

File: skin/src/vue/c-config-menu.vue (hunk 3 above)
Comment (manuelleduc) [reply skin/src/vue/c-config-menu.vue:10]: Icon UI components introduced, see the documentation: https://cristal.xwiki.org/xwiki/bin/view/UIComponents/Vue/Icons/

File: skin/src/vue/c-content.vue (hunk 4 above)
Comment (manuelleduc) [reply skin/src/vue/c-content.vue:119]: ok, we can keep it like that for now.

File: skin/src/vue/c-view.vue (hunk 5 above)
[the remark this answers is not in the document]
Comment (manuelleduc) [reply skin/src/vue/c-view.vue:57]: > I could update the layout to not use a grid layout, but I feel grid gives us more flexibility.

That's fine, but I suggest adding comments explaining the role of rules that are not trivial.

File: skin/src/vue/c-footer.vue [hunk 6]
```diff
@@ -23,7 +23,7 @@
  *
 -->
 <template>
-  <footer id="footer">
+  <footer id="footer" class="footer">
```
manuelleduc, line 26: Out of curiosity, why do you need an additional `footer` class when you already have a `footer` id?
Couldn't the CSS below equality be `#footer { grid-area:footer; }` ?
Also, is the `footer` param of `grid-area` expected to match a `grid-template-areas` definition (I can't find a `grid-template-areas` with a `footer` value)?

tkrieck: Usually I don't like to use IDs on CSS if I have control over the HTML. I prefer to leave IDs reserved for JS functions. The footer layout is an old file and the new layout don't have it, it could be removed.

manuelleduc: ok, thanks for the explanations!

lib/src/c-index.vue
tkrieck: For now, I've moved all component styles inside the components themselves and kept the true global styles on index.vue. 

In index.css I have placed all custom CSS properties on a map 1:1 with shoelace but with a "cr" prefix, so in the future, should we support a new DS, these mappings can be reassigned. Let me know if there's a more appropriate file for these customizations.

skin/src/vue/c-config-menu.vue
tkrieck: Thanks, I forgot to comment here. 

I changed all instances of icons outside of Shoelace to the new c-icon component. However, we still have a bug (at least, I think it is a bug) with slots for icons of Shoelace itself. Example, the add button besides the breadcrumb should have a "+" icon, but the icon does not appear. Maybe it has something to do with this: https://stackoverflow.com/questions/76446464/how-to-expose-shoelace-icons-in-a-vite-app

core/icons/src/vue/c-icon.vue
```diff
@@ -3,17 +3,21 @@ import "bootstrap-icons/font/bootstrap-icons.css";
 
 defineProps<{
   name: string;
+  size: string;
```
manuelleduc, line 6: I suggest making this size optional, as afaics many uses of `<c-icon>` on templates does not define it.
Also, since it can be undefined, this needs to be taken into account.
When undefined, the classes generated on the span are `cr-icon bi-NAME undefined`, where we would like `cr-icon bi-NAME`.
```suggestion
  size?: string;
```

core/icons/src/vue/c-icon.vue
```diff
@@ -3,17 +3,21 @@ import "bootstrap-icons/font/bootstrap-icons.css";
...
 }>();
 // TODO: currently names are bound to bootstrap class names. We'll need to add
 // an indirection as soon as we want to support several icon sets.
 </script>
 
 <template>
-  <span :class="`bi-${name}`"></span>
+  <span :class="`cr-icon bi-${name} ${size}`"></span>
```
manuelleduc, line 13: What's the `cr-icon` for? 
I can see it used in [c-content.vue](https://github.com/xwiki-contrib/cristal/pull/118/files#diff-90174661011634404d087d75ca18635966e4a6ab666480897464453fac3e58e1R326) but since it's in a scoped style block, it is ignored.

core/icons/src/vue/c-icon.vue
```diff
@@ -3,17 +3,21 @@ import "bootstrap-icons/font/bootstrap-icons.css";
...
 </template>
 
 <style scoped>
 span {
-  font-size: 24px;
+  font-size: 1.3rem;
+}
+span.small {
```
manuelleduc, line 20: Is there two sizes supported currently: "undefined", and small?

core/icons/src/vue/c-icon.vue
tkrieck: It should be a generic class to target Cristal icons used in the components layout. But now I remember this class will be suffixed right? Maybe I need to target the component itself (c-icon)

core/icons/src/vue/c-icon.vue
tkrieck: Ah ok, got it.

PR #171, opened 2024-04-18
CRISTAL-165: Initialize the main layout

tkrieck:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-165

# Changes

* Changes in HTML structure for multiple DS
* Provide independend layout for edit mode

## Description

* This Pr is a continuation of a [previous PR](https://github.com/xwiki-contrib/cristal/pull/118) that dealt with the main layout of Cristal. 

## Clarifications

-

# Screenshots & Video
Screenshot 2024-04-18 at 06 40 58

# Executed Tests

-

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


core/icons/src/vue/c-icon.vue
```diff
@@ -37,7 +37,7 @@ const classes = computed(() => {
 </script>
 
 <template>
-  <span :class="classes"></span>
+  <span :class="'cr-icon ' + classes"></span>
```
manuelleduc, line 40: Instead of relying on concatenation, it's cleaner to move `cr-icon` to the classes [line 28](https://github.com/xwiki-contrib/cristal/pull/171/files#diff-ff49e796119f8e661c1ab1642f7d78a4c6f3d49d34376505441466a1442d3e19R28).

manuelleduc: the `computed` method you can see line 27 is a special Vue method that captures reactive variables used in the body of the lamba it takes in parameter.
The value of the computed is only updated when one of the captures reactive variable is itself updated.
See https://vuejs.org/guide/essentials/computed.html for more details.

ds/shoelace/src/vue/css/style.css
```diff
@@ -0,0 +1,330 @@
+:root {
+
+
+   --cr-sizes-max-page-width: 960px;
```
manuelleduc, line 4: Dumb question, but I'm assuming `cr` as in **CR**istal?

editors/prosemirror/src/vue/c-edit-prosemirror.vue
```diff
@@ -122,35 +122,68 @@ const submit = async () => {
 </script>
 
 <template>
-  <div v-if="loading">
+  <div v-if="loading" class="content-loading">
     <!-- TODO: provide a proposer loading UI. -->
-    LOADING
+
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 200"><circle fill="#B9B9B9" stroke="#B9B9B9" stroke-width="6" r="15" cx="40" cy="100"><animate attributeName="opacity" calcMode="spline" dur="2" values="1;0;1;" keySplines=".5 0 .5 1;.5 0 .5 1" repeatCount="indefinite" begin="-.4"></animate></circle><circle fill="#B9B9B9" stroke="#B9B9B9" stroke-width="6" r="15" cx="100" cy="100"><animate attributeName="opacity" calcMode="spline" dur="2" values="1;0;1;" keySplines=".5 0 .5 1;.5 0 .5 1" repeatCount="indefinite" begin="-.2"></animate></circle><circle fill="#B9B9B9" stroke="#B9B9B9" stroke-width="6" r="15" cx="160" cy="100"><animate attributeName="opacity" calcMode="spline" dur="2" values="1;0;1;" keySplines=".5 0 .5 1;.5 0 .5 1" repeatCount="indefinite" begin="0"></animate></circle></svg>
```
manuelleduc, line 128: Can you add a todo before the svg element?
1. We are going to have loading spinner in several places, so better make this reusable
2. I think that this spinner should be design system specific, so we'll need to introduce a `x-spinner` component

editors/prosemirror/src/vue/c-edit-prosemirror.vue
```diff
@@ -122,35 +122,68 @@ const submit = async () => {
...
+  gap: var(--cr-spacing-x-small);
+  padding: var(--cr-spacing-x-small) var(--cr-spacing-x-small);
+  background: var(--cr-color-neutral-100);
+  width: var(--cr-spacing-max-page);
+  margin: var(--cr-spacing-x-small) auto;
+  border-radius: var(--cr-input-border-radius-medium);
+  max-width: var(--cr-sizes-max-page-width);
+  width: 100%;
+}
+:global(.ProseMirror-menubar) {
```
manuelleduc, line 181: Global does not look good here as it could apply outside this component. I think `:deep` (https://vuejs.org/api/sfc-css-features.html#deep-selectors) is more relevant here as it will only apply to this component of it's inner dom.

skin/src/vue/c-content.vue
```diff
@@ -99,11 +99,12 @@ onUpdated(() => {
 });
 </script>
 <template>
-  <div v-if="loading">
+  <div v-if="loading" class="content-loading">
```
manuelleduc, line 102: Note for later: we are seeing a pattern emerging where a screen has 3 states (loading, error, default).
It's probably possible to factors those in a reusable component.

skin/src/vue/c-content.vue
```diff
@@ -136,20 +131,37 @@ onUpdated(() => {
               <span class="doc-info-user-info">
                 User Name edited on 12/12/2024 at 12:00
               </span>
+
               <div class="doc-info-actions">
                 <div class="info-action like">
-                  <c-icon name="heart"></c-icon>
+                  <c-icon name="heart" class="small"></c-icon>
```
manuelleduc, line 137: hum, now we have two ways to define the size of an icons, using the `size` props, or by adding a class.
Do they fulfil different development cases, or is it two ways to do the same thing?

skin/src/vue/c-sidebar.vue
```diff
@@ -33,12 +33,18 @@ const logo = xlogo;
 </script>
 <template>
   <div class="collapsed-sidebar">
+    <!-- This icon should add the class .is-visible to #sidebar.-->
```
manuelleduc, line 36: I'm not sure to understand that comment.

ds/shoelace/src/vue/css/style.css
tkrieck: Indeed =)

skin/src/vue/c-sidebar.vue
manuelleduc: Ah right, you mean that in order to make this panel reactive, we need to add the is-visible class from javascript?

tkrieck: I will reframe the comment. But what I want to say is when the user click this icon (visible only on mobile) a function should add the class `is-visible` to the element #sidebar, this way the CSS will make the sidebar visible again. There's an icon on the sidebar itself, visible only on mobile, that should care to remove `is-visible` again and collapse the sidebar

ds/shoelace/src/vue/css/style.css
manuelleduc: We are going to need to document all of those in a wiki page.
Somebody implementing a new design system must be able to understand them.

skin/src/vue/c-content.vue
tkrieck: Yes, you're right, I forgot about the new method.

core/icons/src/vue/c-icon.vue
tkrieck: Adjusted as instructed.

editors/prosemirror/src/vue/c-edit-prosemirror.vue
tkrieck: Done

skin/src/vue/c-content.vue
tkrieck: Agreed, I will leave this up to you.

skin/src/vue/c-sidebar.vue
tkrieck: I've left a few comments on each icon that should have an action regarding the collapse and hiding of the sidebar. Basically, everything will be done by two classes, one in the element wrapper (`.wrapper .sidebar-is-collapsed`) and another one on the sidebar itself (`#sidebar .is-visible`). I've tried to explain as much as possible in the comments, but let me know if you need something else.

core/icons/src/vue/c-icon.vue
```diff
@@ -25,7 +25,7 @@ function sizeToClass(size: Size): string | undefined {
 }
 
 const classes = computed(() => {
-  const ret = [`bi-${props.name}`];
+  const ret = [`cr-icon bi-${props.name}`];
```
manuelleduc, line 28:
```suggestion
  const ret = ["cr-icon", `bi-${props.name}`];
```

lib/src/index.css
```diff
@@ -23,328 +23,144 @@
  * @license    http://opensource.org/licenses/AGPL-3.0 AGPL-3.0
  *
  **/
- :root {
+ 
+
+ html,
```
manuelleduc, line 28:
```suggestion
html,
```

PR #253, opened 2024-05-23
CRISTAL-179: Make the left panel reactive

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-179

# Changes

## Description

* Left sidebar is now reactive and auto-closes on mobile
* Left sidebar can be set to float, pinned, or closed, and the states are persisted
* Left sidebar can be resized in desktop mode and the changes are persisted

## Clarifications

* This PR adds two reusable composables: one to check the current viewport (mobile/desktop) and one to get the mouse cursor position.
* CSS was slightly altered to make sure that all the buttons to control the left sidebar (collapse, pin and close) appear at the same level.
* Note that, on the video, there is no "resize cursor" only because my cursor theme is currently broken.

# Screenshots & Video

https://github.com/xwiki-contrib/cristal/assets/7363343/c993f41d-5a3b-45c3-b9d4-356e1be85fdf

# Executed Tests

Everything was tested manually.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


skin/src/composables/mouse.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ **/
+import { Ref, onMounted, onUnmounted, ref } from "vue";
+
+interface MouseState {
+  readonly x: Ref<number>;
+  readonly y: Ref<number>;
+}
+
+export function useMouse(): MouseState {
```
manuelleduc, line 32: wdyt of something more descriptive like `useMouseCoordinates` (keep the `use` prefix from Vue naming conventions)?

lib/src/index.css
```diff
@@ -101,6 +101,7 @@ main {
 .left-sidebar {
   width: var(--cr-sizes-left-sidebar-width);
   min-width: var(--cr-sizes-left-sidebar-min-width);
+  max-width: 100%;
```
manuelleduc, line 104: ping @tkrieck for a second look on the CSS changes :)

pjeanjean: Oh yeah, did not mention this specific change. I added that one to make sure that the resized sidebar could not be larger than the viewport, if the user later resizes their window.

skin/src/composables/viewport.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export enum ViewportType {
+  Mobile,
+  Desktop,
+}
+
+export function useViewportType(): Ref<ViewportType> {
+  const viewportWidth: Ref<number> = ref(window.innerWidth);
+  const viewportSize: Ref<ViewportType> = computed(() => {
+    if (viewportWidth.value <= 600) {
```
manuelleduc, line 35: @tkrieck do we have a definition of the viewport size somewhere else on the CSS side?
Trying to avoid having two conflicting definitions.

skin/src/vue/c-sidebar.vue
```diff
@@ -23,33 +23,122 @@
...
 import CHelp from "./c-help.vue";
 import xlogo from "../images/xwiki-logo-color.svg";
 import xavatarImg from "../images/no-one.svg";
 import { CIcon } from "@xwiki/cristal-icons";
+import { useMouse } from "../composables/mouse";
+import { ViewportType, useViewportType } from "../composables/viewport";
 
 const logo = xlogo;
 const avImg = xavatarImg;
+const viewportSize: Ref<ViewportType> = useViewportType();
```
manuelleduc, line 39: I feel like `viewportType` is more accurate here (and also in `useViewportType` implementation)
```suggestion
const viewportType: Ref<ViewportType> = useViewportType();
```

skin/src/composables/mouse.ts
manuelleduc: Also, deserve some documentation:
- textual description how to use this method
- `@since 0.8` annotation

skin/src/composables/viewport.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ **/
+import { Ref, computed, onMounted, onUnmounted, ref } from "vue";
+
+export enum ViewportType {
+  Mobile,
+  Desktop,
+}
+
+export function useViewportType(): Ref<ViewportType> {
```
manuelleduc, line 32: Deserve some documentation:
- textual description how to use this method
- `@since 0.8` annotation

skin/src/composables/viewport.ts
```diff
@@ -0,0 +1,50 @@
...
+  const viewportWidth: Ref<number> = ref(window.innerWidth);
```
manuelleduc, line 33: Also, a small improvement but I think it is wise to minimize the number of declared reactive variables.
I suggest refactoring the code to directly update `viewportSize` in `onResize`.

skin/src/vue/c-sidebar.vue
```diff
@@ -23,33 +23,122 @@
...
+const { x }: { x: Ref<number> } = useMouse();
```
manuelleduc, line 40: The type of `x` can be inferred from the right-hand side return type, no need to redefine it on the left-hand side.
```suggestion
const { x } = useMouse();
```

skin/src/composables/viewport.ts
tkrieck: Not that I remember, I tried to keep everything centralized on the mobile section.

pjeanjean: I wonder if there is a clean way to store the definition "globally" through Vue, and inject it in the CSS whenever we need it.

tkrieck:
> I added a few improvements and fixed the collapse/shrink behavior (that was basically inverted). It should be a lot more ergonomic, and now I think that the expand on hover is no longer required.

I checked the last version and everything seems to be fine. Indeed, hover will not be needed

ds/shoelace/src/vue/css/style.css
```diff
@@ -285,4 +285,4 @@
   --cr-z-index-toast: var(--sl-z-index-toast);
   --cr-z-index-tooltip: var(--sl-z-index-tooltip);
 
-}
\ No newline at end of file
+}
```
manuelleduc, line 288: If it's the only change of that file, we can do without it

skin/src/composables/viewport.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Returns a reactive state that contains the type of the current viewport.
+ * @return a reactive state with either of these values:
+ *  * ViewportType.Mobile
+ *  * ViewportType.Desktop
+ *
+ * @since 0.8
+ **/
```
manuelleduc, line 39: WDYT if mentioning the 600px constant in here?

lib/src/index.css
tkrieck: Seems like a good failsafe. As @pjeanjean mentioned.

PR #277, opened 2024-06-07
CRISTAL-203: Tables are missing CSS

tkrieck:
* Created basic styles for the tables

# Jira URL

https://jira.xwiki.org/browse/CRISTAL-203

# Changes

## Description

* Creates the basic styles for tables inside documents

## Clarifications

* For now, these styles are done with the ```:global``` tag because tables that come from the XWiki backend don't have any id, just basic HTML.

# Screenshots & Video

Screenshot 2024-06-07 at 10 12 55

# Executed Tests

-

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/skin/src/vue/c-content.vue
+++ b/skin/src/vue/c-content.vue
@@ -213,6 +213,45 @@ onUpdated(() => {
   justify-content: center;
 }
 
+/*TABLE STYLES*/
+/*TODO: Check a better way to write these styles without the global tag. Currently impossible to use :deep because the html inside the document content is not assigned an ID */
+
+:global(.document-content table) {
+  max-width: 100%;
+  width: 100%;
+  border: none;
+  border-collapse: collapse;
+  color: inherit;
+  margin-bottom: var(--cr-spacing-small);
+}
+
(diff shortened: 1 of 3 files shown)


skin/src/vue/c-content.vue
```diff
@@ -213,6 +213,45 @@ onUpdated(() => {
...
+  margin-bottom: var(--cr-spacing-small);
+}
+
+:global(.document-content table tr) {
+  border-bottom: solid 1px var(--cr-input-border-color);
+}
+
+:global(.document-content table th) {
+  font-weight: var(--cr-font-weight-bold);
+  text-align: left;
```
michitux, line 234:
```suggestion
  text-align: start;
```

Cristal shouldn't repeat the mistake of XWiki to not write styles with RTL languages in mind, in particular as there are easy ways to keep the CSS generic like using `start` instead of `left` for `text-align`.

tkrieck: Agreed, thanks for the heads up. I will make a pass on the source for other instances where this can be applied.

PR #301, opened 2024-06-25
CRISTAL-212: Make embedding less prone to CSS conflicts

tkrieck:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-212

# Changes

## Description

* Changed the CSS architecture to use grid in place of fixed positioning. This should enable Cristal to be embedded in Nextcloud without problems in layout disposition.
* Changed classes and IDs names to make conflicts with embedding apps less prevalent.

## Clarifications

* The changes brought by this PR, unfortunately, disables the nice sidebar animations that we had before. I tried as much as I can to bring them back but for now, they remain off. 
* I will still try to animate elements, but I decided to make this work a PR to enable further testing on the layout changes.

# Screenshots & Video

-

# Executed Tests

-

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


electron/renderer/index.html
```diff
@@ -1,12 +1,12 @@
 <!doctype html>
-<html>
+<html class="xw-cristal">
```
manuelleduc, line 2: I assume this is a special class when one wants to take over the full windows?

lib/src/index.css
```diff
@@ -16,214 +16,9 @@
...
-
-  .wrapper .sidebar-collapse-controls .pin-sidebar {
-    display: none;
-  }
-
-  .resize-handle {
-    display: none;
-  }
-}
+ *
```
manuelleduc, line 19: Why did had to move everything to c-view?

web/index.html
```diff
@@ -16,6 +16,6 @@
         <style type="css" src='./src/index.css'></style>
 </head>
 <body>
-	<div id="app"></div>
+	<div id="xwCristalApp"></div>
```
manuelleduc, line 19: I assume this is to make it less likely to conflict with another id from the service where we embed Cristal?
If that's the case, it's ok for now but we'll need to make this id configurable later.

lib/src/index.css
tkrieck: To maintain it inside Vue scoping rules. This way an ID is assigned by Vue when generating the CSS for the browser helping with isolation of rules.

Screenshot 2024-06-26 at 09 33 03

I chose `c-view` because it seemed to be the most generic one. We could separate the rules to be used inside each component in which a rule is needed, but for generic layout rules I feel this would make maintenance too complicated.

electron/renderer/index.html
tkrieck: This will make it easier when we need to specify a CSS class only for Cristal. Especially in cases where we have more generic class names. For example:` .content .p` could interfere with the embedding app so instead we can write `.xw-cristal .content .p`. This is a generic example, by the way, I don't remember seeing something like this outside of Vue scoping in our code.

web/index.html
tkrieck: Correct, `#app` is very commonly used.

editors/tiptap/src/vue/c-edit-tiptap.vue
```diff
@@ -215,43 +215,39 @@ onUpdated(() => loadEditor(currentPage.value!));
...
-            :editor="editor"
-          ></c-tiptap-bubble-menu>
-          <editor-content :editor="editor" class="document-content editor" />
-        </div>
+  <div v-show="!loading && !error" class="content">
+    <input
+      v-model="title"
+      type="text"
+      :placeholder="titlePlaceholder"
+      class="document-title"
```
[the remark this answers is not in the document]
tkrieck: Yes, that was the problem. I have no strong preferences besides that shorter names are quicker to write and read. If you have a strong preference for longer, more descriptive class names, I can change back to `document-` prefix

electron/renderer/index.html
tkrieck: > Where would this CSS class be put when Cristal is embedded?

We are embedding on Nextcloud only for now, I am not sure if we'll eventually embed on other systems as well. On Nextcloud embedding, we have control over the tag that will make the embedding, via templates. In our **current** nextcloud template, this class is declared as `#app`. I am proposing to change it to `#xwCristalApp` to make it more specific. https://github.com/xwiki-contrib/cristal/pull/301/files/4147d616e85546c534943cc81eef54db04db4bf9#r1654690536

One thing I forgot to mention is that I need this class to target the upmost available tag under our control to set the starting CSS properties and `#xwCristalApp` should be inside of it. That's because I can't simply target `html` on our CSS, because that could break Nextcloud's CSS. You can check the CSS rules in `c-view.vue`, line 92. **Important**: I made a mistake there, addressing it to `html.xw-cristal` it should only target `.xw-cristal`, this should be fixed on the next commit.

> It appears to me that using only one is enough. In which case I'm leaning towards using the CSS class

I agree on general terms, but as I commented above, I really need to target the topmost HTML tag that we have access to, no matter the context.

tkrieck: > But if we do this, then do we need both an id xwCristalApp and a CSS class xw-cristal? It appears to me that using only one is enough. In which case I'm leaning towards using the CSS class (even though having two instances of the Cristal application embedded on the same page is not expected, is it?).

After giving this some thought and testing, you are right. I can get both on the same element, the ID is now used only for mounting the app via TS and the CSS class for styling and positioning (as it should be). The CSS is now cleaner as well, and my previous comment is somewhat obsolete, so all in all it was a good change, thanks!

lib/src/index.css
```diff
@@ -22,3 +22,8 @@
  * @license    http://opensource.org/licenses/AGPL-3.0 AGPL-3.0
  *
  **/
+ html,
+ body {
+         width: 100%;
+         height: 100%
```
mflorea, line 28: Strange indentation.

mflorea (approved):
Looks good. Thanks!

PR #337, opened 2024-08-23
CRISTAL-201: Introduce a Navigation tree

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-201

# Changes

## Description

* Introduce an API to manage navigation trees: NavigationTreeResolver
* Implement NavigationTreeResolver components for multiple backends:
  * FileSystem
  * GitHub
  * Nextcloud
  * XWiki
* Add navigation tree UI element to Shoelace and Vuetify
* Add a page test for XWiki backend

## Clarifications

The current implementations are fully lazy.

This PR can be improved on a few points, that should probably be addressed in future issues:
- Currently, when expanding an item, there is no check on the great grandchildren. This means that children can, in some instances, wrongly display as expandable event though they do not have any children (*e.g.*, this happens with the XWiki backend, since Attachments and Translations are considered children for the `DocumentTree` sheet, but I filter them in Cristal).
- This uses the UI elements offered by Shoelace and Vuetify without any visual modification. We probably need to do work on this because Vuetify's tree view (in particular) looks very out of place inside the sidebar.
- Support for icons will need to be added whenever we have final design specifications on this.

# Screenshots & Video

XWiki Shoelace
image

GitHub Shoelace
image

FileSystem Shoelace
image

Nextcloud Shoelace
image

XWiki Vuetify
image

# Executed Tests

Everything was tested manually, current tests work and the new page test also passes properly.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


ds/shoelace/src/vue/x-navigation-tree.vue
```diff
@@ -0,0 +1,81 @@
...
+  rootNodes.value.push(...(await props.treeResolver.getNavigationTree("")));
+});
+
+function lazyLoadChildren(id: string) {
+  return async (event: Event) => {
+    const lazyItem = event.target! as Element;
+    const childNodes = await props.treeResolver.getNavigationTree(id);
+    for (const child of childNodes) {
+      const treeItem = document.createElement("sl-tree-item");
+      treeItem.innerHTML = `<a href="${child.url}">${child.label}</a>`;
```
mflorea, line 44: The label needs to be escaped as it can contain HTML special characters.

ds/shoelace/src/vue/x-navigation-tree.vue
```diff
@@ -0,0 +1,81 @@
...
+    lazyItem.removeAttribute("lazy");
+  };
+}
+</script>
+
+<template>
+  <sl-tree>
+    <sl-tree-item
+      v-for="item in rootNodes"
+      :key="item.label"
```
mflorea, line 72: The label can be duplicated no? Isn't it better to use the node id as key?

mflorea (approved):
Looks good! Thanks, @pjeanjean !

PR #375, opened 2024-10-09
CRISTAL-244: Clicking outside the text in the navigation tree nodes does nothing

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-244

# Changes

## Description

Vuetify's "activation" and Shoelace's "selection" are now fully handled manually to only have a single node selected: the current page (or clicked link if a clickAction is present).
Confusing hover effects have also been disabled.

## Clarifications

~~In this PR, Vuetify's navigation tree was also made more compact to align a bit more with Shoelace's one.~~

# Executed Tests

Test suite was run fully.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/ds/shoelace/src/utils/navigation-tree-selection.ts
+++ b/ds/shoelace/src/utils/navigation-tree-selection.ts
@@ -1,0 +1,29 @@
+import { type Ref } from "vue";
+import type SlTreeItem from "@shoelace-style/shoelace/dist/components/tree-item/tree-item";
+
+export class NavigationTreeSelection {
+  private selectedTreeItem: SlTreeItem | undefined = undefined;
+  private treeItems: Ref<Map<string, SlTreeItem>>;
+
+  constructor(treeItems: Ref<Map<string, SlTreeItem>>) {
+    this.treeItems = treeItems;
+  }
+
+  public updateSelection(id: string) {
+    if (this.selectedTreeItem) {
+      this.selectedTreeItem.selected = false;
+    }
+    this.selectedTreeItem = this.treeItems.value.get(id);
+    this.selectedTreeItem!.selected = true;
+  }
+
+  public onSelectionChange(event: unknown) {
+    // We don't want users to manually select a node, so we undo any change.
+    (
+      event as { detail: { selection: SlTreeItem[] } }
+    ).detail.selection[0].selected = false;
+    if (this.selectedTreeItem) {
+      this.selectedTreeItem!.selected = true;
+    }
+  }
+}
--- a/ds/shoelace/src/utils/navigation-tree-selection.ts
+++ b/ds/shoelace/src/utils/navigation-tree-selection.ts
@@ -17,6 +17,10 @@ export class NavigationTreeSelection {
     this.selectedTreeItem!.selected = true;
   }
 
+  public getSelection(): SlTreeItem | undefined {
+    return this.selectedTreeItem;
+  }
+
   public onSelectionChange(event: unknown) {
     // We don't want users to manually select a node, so we undo any change.
     (
--- a/ds/shoelace/src/utils/navigation-tree-selection.ts
+++ b/ds/shoelace/src/utils/navigation-tree-selection.ts
@@ -1,6 +1,31 @@
+/*
+ * See the LICENSE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
 import { type Ref } from "vue";
 import type SlTreeItem from "@shoelace-style/shoelace/dist/components/tree-item/tree-item";
 
+/**
+ * Utility class to manage the selected node in a Navigation Tree for Shoelace ds.
+ *
+ * @since 0.11
+ **/
 export class NavigationTreeSelection {
   private selectedTreeItem: SlTreeItem | undefined = undefined;
   private treeItems: Ref<Map<string, SlTreeItem>>;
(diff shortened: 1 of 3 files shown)


manuelleduc:
Thanks for working on this @pjeanjean.
First UI related remark: when hovering the node, the cursor becomes a "hand", signaling an action on click.
But, nothing happens unless the click is done on the label.

ds/vuetify/src/vue/x-navigation-tree.vue
```diff
@@ -28,8 +28,9 @@ import type {
 
 type TreeItem = {
   id: string;
+  href: undefined;
```
manuelleduc, line 31: Might worth a comment if the intent is indeed to force href to always be undefined.

ds/vuetify/src/vue/x-navigation-tree.vue
```diff
@@ -128,38 +131,54 @@ async function lazyLoadChildren(item: unknown) {
...
 
 <template>
   <v-treeview
     ref="tree"
-    v-model:activated="activatedNodes"
     v-model:opened="expandedNodes"
+    :activated="activatedNodes"
     :items="rootNodes"
     :load-children="lazyLoadChildren"
     activatable
```
manuelleduc, line 152: I might be missing something. Isn't just removing this option enough to fix the issue? See https://vuetifyjs.com/en/components/treeview/#activatable

ds/vuetify/src/vue/x-navigation-tree.vue
```diff
@@ -128,38 +131,54 @@ async function lazyLoadChildren(item: unknown) {
...
-    active-strategy="single-independent"
+    active-strategy="independent"
+    density="compact"
```
manuelleduc, line 154: We do tou change the density? It seems unrelated to the current issue.

ds/vuetify/src/vue/x-navigation-tree.vue
```diff
@@ -128,38 +131,54 @@ async function lazyLoadChildren(item: unknown) {
...
       >
-      <a v-else :href="item.href">{{ item.title }}</a>
+      <a v-else :href="item.url">{{ item.title }}</a>
     </template>
   </v-treeview>
 </template>
 
-<style scoped></style>
+<style scoped>
+:deep(.v-list-item__overlay) {
```
manuelleduc, line 181: This deserve a comment.

ds/shoelace/src/vue/x-navigation-tree.vue
```diff
@@ -91,6 +87,14 @@ function expandTree() {
   }
 }
 
+function updateSelection(id: string) {
```
manuelleduc, line 90: General comment on this page changes. Wouldn't it be simpler to not wrap the labels with an a element and to navigate on selection change?

ds/vuetify/src/vue/x-navigation-tree.vue
pjeanjean: Well, the point was to fix [this remark](https://github.com/xwiki-contrib/cristal/pull/375#issuecomment-2402189272), and it seemed to work when I tested it but now I reproduce it again. This seems too inconsistent so I will revert this part and use a CSS rule instead.

ds/vuetify/src/vue/x-navigation-tree.vue
pjeanjean: Removing activatable has the side-effect of making non-leaf nodes impossible to activate (no matter the combination of options used), and I want to activate them on page change to display the current page.

ds/shoelace/src/vue/x-navigation-tree.vue
pjeanjean: I'm not sure I understand what you are suggesting here.

ds/vuetify/src/vue/x-navigation-tree.vue
manuelleduc: Hum, the point of having several design systems is precisely to make them different ;)
We need to discuss this separately with @tkrieck if we want to change it.

ds/shoelace/src/vue/x-navigation-tree.vue
```diff
@@ -165,12 +189,20 @@ function lazyLoadChildren(id: string) {
       <a
         v-if="props.clickAction"
         :href="item.url"
-        @click.prevent="clickAction!(item)"
+        @click.prevent="
+          updateSelection(item.id);
```
manuelleduc, line 193: Looks duplicated with line 151. Wdyt of introducing a common method?

ds/shoelace/src/vue/x-navigation-tree.vue
```diff
@@ -165,12 +189,20 @@ function lazyLoadChildren(id: string) {
...
         >{{ item.label }}</a
       >
       <a v-else :href="item.url">{{ item.label }}</a>
     </sl-tree-item>
   </sl-tree>
 </template>
 
-<style scoped></style>
+<style scoped>
+/* Disable hand cursor on items. */
```
manuelleduc, line 204: Can you explain *why* we need to disable the hand cursor on item?

ds/shoelace/src/vue/x-navigation-tree.vue
```diff
@@ -106,8 +124,13 @@ function onMutation(mutationList: Array<MutationRecord>) {
...
 
 function onSelectionChange(event: unknown) {
-  selectedTreeItem = (event as { detail: { selection: SlTreeItem[] } }).detail
-    .selection[0];
+  // We don't want users to manually select a node, so we undo any change.
+  (
+    event as { detail: { selection: SlTreeItem[] } }
+  ).detail.selection[0].selected = false;
+  if (selectedTreeItem) {
+    selectedTreeItem!.selected = true;
```
manuelleduc, line 132: wdyt of grouping selectedTreeItem and all the methods that are manipulating it in a composable (are at least encapsualte all this)?

ds/vuetify/src/vue/x-navigation-tree.vue
```diff
@@ -162,4 +184,13 @@ async function lazyLoadChildren(item: unknown) {
   </v-treeview>
 </template>
 
-<style scoped></style>
+<style scoped>
```
manuelleduc, line 187: Here too, can you explain also why we need to disable those elements?

ds/shoelace/src/utils/navigation-tree-selection.ts
```diff
@@ -0,0 +1,33 @@
+import { type Ref } from "vue";
```
manuelleduc, line 1: Missing license header

ds/shoelace/src/utils/navigation-tree-selection.ts
```diff
@@ -0,0 +1,33 @@
...
+import type SlTreeItem from "@shoelace-style/shoelace/dist/components/tree-item/tree-item";
+
+export class NavigationTreeSelection {
```
manuelleduc, line 4: Would be good to have a `since` annotation for traceability and a some javadoc.

manuelleduc (changes requested):
Two small changes and we are good to merge :)

PR #399, opened 2024-10-22
CRISTAL-82: Deleting a page is possible

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-82

# Changes

## Description

* This adds support for page actions and categories as UI Extensions.
* A basic support to create and display alerts on the UI was also added through the service `AlertsService`.
* The operation `delete(page)` has been implemented for `XWiki`, `Nextcloud` and `FileSystem` backends.
* `PageHierarchyItem` were extended with the property `pageId`, which is useful to list and access parent pages outside of just the breadcrumb.
* The UI now displays a `Manage` category with a `Delete` action as page actions.
* A few components were also improved slightly, e.g., the variant `text` is now usable for buttons.

## Clarifications

This PR ended up bringing a lot of new features to implement the following process:
- A button Delete is available when clicking the 3 vertical dots in page actions
- Clicking the button shows a confirm dialog
- After the deletion is confirmed, the user is redirected to the parent page and an alert notifies him that the page as successfully deleted

While working on this, the login feature was merged to main which made navigation trees noticeably larger.
I noticed that the navigation tree for XWiki did not display more than 15 items at a time because the API used had pagination that was not handled, so this PR also includes a fix for that.

# Screenshots & Video

image
image
image

# Executed Tests

The test suite was run and the features were tested manually.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


api/src/api/storage.ts
```diff
@@ -94,4 +94,14 @@ export interface Storage {
    * @since 0.9
    */
   saveAttachments(page: string, files: File[]): Promise<unknown>;
+
+  /**
+   * Delete a page.
+   *
+   * @param page the page to delete
+   * @return a promise on the delete is done
```
manuelleduc, line 102: on/when

pnpm-workspace.yaml
```diff
@@ -27,6 +28,7 @@ packages:
   - "core/icons"
   - "core/info-actions/*"
   - "core/link-suggest/*"
+  - "core/page-actions/*"
```
manuelleduc, line 31: This should be one line below, we order the packages by alphabetical order.

core/page-actions/page-actions-ui/src/vue/PageActionsCategory.vue
```diff
@@ -0,0 +1,63 @@
...
+    actions.value.push({
+      action: currAction,
+      component: await currAction.component(),
+    });
+  }
+});
+</script>
+<template>
+  <x-menu-label>{{ category.title }}</x-menu-label>
+  <x-menu-item v-for="a of actions" :key="a.action.id" :value="a.action.id">
```
manuelleduc, line 54: WDYT of renaming 'a' to 'action'?

manuelleduc (commented):
First batch of small remarks from a first quick pass.
Also, from testing a running version with the XWiki backend:
- when the user is Guest, deleting the page make the page reload instead of showing an error (because the user does not have delete rights)
- when the user has delete rights, I have a info message but the page isn't actually removed (I get `Authentication failure with login [null]` on XWiki's logs)

core/alerts/alerts-ui/src/vue/AlertsToasts.vue
```diff
@@ -0,0 +1,71 @@
...
+<template>
+  <div class="wrapper">
+    <XAlert
+      v-for="alert of alerts"
+      :key="alert.id"
+      :type="alert.type"
+      :description="alert.message"
+      :actions="alert.actions"
+      @update:model-value="
+        ($event: boolean) => {
```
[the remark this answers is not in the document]
pjeanjean: This is the behavior of Vue's `update:modelValue`: the only value sent to the listener is the new model's value.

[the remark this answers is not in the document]
manuelleduc: ok, so maybe `$event` would be better named differently? If I see `$event` I'm always expecting it to have the `Event` type.

ds/shoelace/src/vue/x-alert.vue
```diff
@@ -17,49 +17,62 @@ License along with this software; if not, write to the Free
...
-  title: { type: String, required: true },
-  type: {
-    type: String as PropType<Types>,
-    required: true,
-  },
-  description: { type: String, required: true },
-});
+const { type } = defineProps<AlertProps>();
 
 const variant = computed(() => {
```
manuelleduc, line 29: By using [destructuring assignments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) you can avoid the intermediate `variant` constant and avoid the odd `variant.variant` and `variant.icon` in the template.
```suggestion
const {variant, icon} = computed(() => {
```

pjeanjean: I tried but TypeScript complains when trying to destructure ComputedRefs

manuelleduc (changes requested):
One small change and we are good to merge.

ds/shoelace/src/vue/x-alert.vue
manuelleduc: ok, there is maybe something to be done with `toRef` but not sure if it would improve readability.
Let's ignore that point for now.

PR #418, opened 2024-11-08
CRISTAL-142: View history if the backend supports it

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-142

# Changes

## Description

* This PR add the components `PageRevisionManager` (backend-dependant) and `PageRevisionManagerProvider` to get the right instance of the former. Implementation for the XWiki and GitHub backends are included.
* A new History extra tab will be displayed, listing the existing revisions for a given document.
* Clicking the link of a revision displays it, and edit operations are temporarily disabled. An alert explaining this is displayed, with a link to reload the latest revision of the document.
* A new utility module, `cristal-xwiki-utils`, was added to factorize operations made for multiple XWiki components.
* A UI test checks that the history is properly displayed, and that links do update the content.

## Clarifications

* The current version does not handle pagination and does not permit to delete a revision, see https://jira.xwiki.org/browse/CRISTAL-291 and https://jira.xwiki.org/browse/CRISTAL-292. These features are currently missing from XWiki's REST API: https://jira.xwiki.org/browse/XWIKI-22645 and https://jira.xwiki.org/browse/XWIKI-22647
* For XWiki, only the user ids are displayed for the creator of the revision. This should change when Cristal has proper support for users, which is also a requirement for https://jira.xwiki.org/browse/CRISTAL-273
* Comparing and restoring revisions was not part of the MVP but will be done in https://jira.xwiki.org/browse/CRISTAL-290

# Screenshots & Video

image
image

# Executed Tests

The changes were tested manually and through the new UI test. The full test suite was run succesfully.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


api/src/api/storage.ts
```diff
@@ -27,7 +27,11 @@ export interface Storage {
 
   getWikiConfig(): WikiConfig;
 
-  getPageRestURL(page: string, syntax: string): string;
+  getPageRestURL(
+    page: string,
+    syntax: string,
+    revision: string | undefined,
```
manuelleduc, line 33: To be double-checked, but I think this syntax is equivalent, more concise and already present in our code base.
```suggestion
    revision?: string,
```

core/backends/backend-dexie/src/wrappingOfflineStorage.ts
```diff
@@ -72,7 +72,7 @@ export class WrappingOfflineStorage implements WrappingStorage {
   }
 
   public getPageRestURL(page: string, syntax: string): string {
-    return this.storage.getPageRestURL(page, syntax);
+    return this.storage.getPageRestURL(page, syntax, undefined);
```
manuelleduc, line 75: If you use the `?` syntax I'm proposing above, I think you don't have to change this line.

api/src/api/cristalApp.ts
```diff
@@ -84,12 +84,14 @@ export interface CristalApp {
...
    * @param page - a page identifier (e.g., a document reference for the XWiki
    *  backend, or a filename for the filesystem backend)
+   * @param revision - the revision requested, undefined will default to latest
    * @returns the page data, or undefined if the page is not found
    *
    * @since 0.7
    */
   getPage(
     page: string,
+    revision: string | undefined,
```
manuelleduc, line 94: Why not make revision part of the optional options?

api/src/api/storage.ts
```diff
@@ -46,6 +50,7 @@ export interface Storage {
   getPageContent(
     page: string,
     syntax: string,
+    revision: string | undefined,
```
manuelleduc, line 53: Same here, making this optional would avoid changing quite a lot of code elsewhere.
1. undefined last parameter
2. unused parameters prefixed with an underscore

core/backends/backend-xwiki/src/xwikiStorage.ts
```diff
@@ -73,16 +74,23 @@ export class XWikiStorage extends AbstractStorage {
...
       this.wikiConfig.baseURL +
       this.wikiConfig.baseRestURL +
       "&page=" +
       encodeURIComponent(page) +
       "&format=" +
-      syntax
-    );
+      syntax;
+    if (revision) {
+      url += "&revision=" + revision;
```
manuelleduc, line 91: Missing `encodeURIComponent` on revision parameter.
I also think we are reaching the point where using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams would be cleaner and clearer.

core/hierarchy/hierarchy-xwiki/package.json
```diff
@@ -20,18 +20,15 @@
   "scripts": {
     "build": "tsc --project tsconfig.json && vite build",
     "clean": "rimraf dist",
-    "lint": "eslint \"./src/**/*.{ts,tsx}\" --max-warnings=0",
-    "test": "vitest --run"
+    "lint": "eslint \"./src/**/*.{ts,tsx}\" --max-warnings=0"
```
manuelleduc, line 23: Why are the tests gone?

core/history/history-api/src/index.ts
```diff
@@ -0,0 +1,77 @@
...
+import type { PageData } from "@xwiki/cristal-api";
+
+/**
+ * Description of a revision for a given page.
+ * @since 0.12
+ */
+type PageRevision = {
+  version: string;
+  date: Date;
+  user: string; // TODO: replace by an actual User instance.
```
manuelleduc, line 30: We have `UserDetails` in `@xwiki/cristal-authentication-api`

core/history/history-api/src/index.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @since 0.12
+ **/
+interface PageRevisionManager {
+  /**
+   * Returns the revisions for a given page.
+   *
+   * @param pageData - the page for which to get the revisions
+   * @returns the page revisions
+   */
+  getRevisions(pageData: PageData): Promise<Array<PageRevision>>;
```
manuelleduc, line 47: I assume the results are not paginated because of https://jira.xwiki.org/browse/XWIKI-22645 ?
It would be nice to still take into account pagination in the API even if some backend does not support it.

core/history/history-default/src/components/componentsInit.ts
```diff
@@ -0,0 +1,72 @@
...
+import type { CristalApp } from "@xwiki/cristal-api";
+
+/**
+ * Default implementation for PageRevisionManager.
+ * This is a fallback that only returns an empty history.
+ *
+ * @since 0.12
+ **/
+@injectable()
+class DefaultPageRevisionManager implements PageRevisionManager {
```
manuelleduc, line 37: Note: I'm sure this is the best approach. What if we want to fully hide the history from the UI when a backend does not support it?
Everything not supported by a given backend should not be part of the UI as a general rule.

core/history/history-default/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+} from "@xwiki/cristal-history-api";
+
+export class ComponentInit {
+  constructor(container: Container) {
+    container
+      .bind<PageRevisionManager>(PageRevisionManagerName)
+      .to(DefaultPageRevisionManager)
+      .inSingletonScope()
+      .whenTargetIsDefault();
+    container
```
manuelleduc, line 39: Note to myself. I think we reached a point where we have enough distinct providers to identify a pattern.
We need to define a reusable generic solution now.

core/history/history-ui/src/vue/HistoryTab.vue
```diff
@@ -0,0 +1,110 @@
...
+      <td>
+        <a :href="revision.url">{{
+          revision.date.toLocaleString(undefined, {
+            dateStyle: "short",
+            timeStyle: "short",
+          })
+        }}</a>
+        {{ t("history.extraTabs.entry.separator") }} {{ revision.user }}
+        <br />
+        {{ revision.comment }}
```
manuelleduc, line 72: A sentence must be a single translation key. Word order can vary from one language to another.
vue-i18n is providing a lot of options for this, just ping me to discuss this.
This is quite easy once you know the right tools.

core/history/history-ui/src/translations.ts
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+import en from "../langs/translation-en.json";
+
+const translations: Record<string, Record<string, string>> = {
+  en,
```
manuelleduc, line 24: Missing fr import?

core/history/history-ui/src/HistoryExtraTab.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import messages from "./translations";
+import HistoryTab from "./vue/HistoryTab.vue";
+import { injectable } from "inversify";
+import { AbstractExtraTab } from "@xwiki/cristal-extra-tabs-api";
+import { Component } from "vue";
+
+@injectable()
+export class HistoryExtraTab extends AbstractExtraTab {
```
manuelleduc, line 28: Related to the comment about hidden the history when the backend supports it. We might be missing a mechanism to exclude an extra tab under some condition.

core/history/history-ui/src/vue/HistoryTab.vue
```diff
@@ -0,0 +1,110 @@
...
+const revisions: Ref<Array<PageRevision>> = ref([]);
+
+onMounted(async () => {
+  if (currentPage.value) {
+    revisions.value = await cristal
+      .getContainer()
+      .get<PageRevisionManagerProvider>("PageRevisionManagerProvider")
+      .get()
+      .getRevisions(currentPage.value);
+  }
```
manuelleduc, line 51: I think we can do better with `documentService.getCurrentDocument()` which return a ref to the current page + computed.
At least we need to check that the history is refreshed on page navigation.

core/history/history-xwiki/src/components/componentsInit.ts
```diff
@@ -0,0 +1,107 @@
...
+          });
+        },
+      );
+      return revisions;
+    } catch (error) {
+      this.logger.error(error);
+      this.logger.debug(
+        `Could not load history for page ${pageData.name}, falling back to default history manager.`,
+      );
+      return this.defaultRevisionManager.getRevisions(pageData);
```
manuelleduc, line 102: It is better to display an actual error to the user in the UI than to log it and show an empty history.

core/xwiki/xwiki-utils/src/index.ts
```diff
@@ -26,7 +26,7 @@ import type { WikiConfig } from "@xwiki/cristal-api";
  * @param wikiConfig - the current wiki configuration
  * @param documentId - the id of the document
  * @returns the crafted URL
- * @since 0.9
+ * @since 0.12
```
manuelleduc, line 29: Why the version change?

web/e2e/main-page.spec.ts
```diff
@@ -144,6 +145,35 @@ configs.forEach(
       )
     });
 
+    test(`[${name}] has working history`, async ({ page }) => {
```
manuelleduc, line 148: Thanks for adding a functional test :)

manuelleduc (commented):
Looks good overall.
I've added a lot of small comments on code style.
I'll do another pass on the running UI asap.

core/hierarchy/hierarchy-xwiki/package.json
pjeanjean: Because they were related to the helper method that I moved to `xwiki/utils`. The tests were moved as well

manuelleduc: Of course, makes sense then.
Couldn't we add other unit tests to this module instead?

core/xwiki/xwiki-utils/src/index.ts
pjeanjean: Since the module is new, it made sense to me to indicate that this specific method can be imported from there since 0.12. Not sure what's the best...

core/history/history-ui/src/translations.ts
pjeanjean: Indeed

core/xwiki/xwiki-utils/src/index.ts
manuelleduc: Ah, right I missed the fact that the file was moved. I think we can keep the original version for moved things.
Though that's something to mention in the release note (assuming the method was publicly exported in the root `index.ts` earlier).

api/src/api/cristalApp.ts
```diff
@@ -84,13 +84,14 @@ export interface CristalApp {
...
    *  backend, or a filename for the filesystem backend)
+   * @param revision - the revision requested, undefined will default to latest
    * @returns the page data, or undefined if the page is not found
    *
    * @since 0.7
    */
   getPage(
     page: string,
-    options?: { requeue: boolean },
+    options?: { requeue: boolean; revision?: string },
```
manuelleduc, line 94: Shouldn't `requeue` be option here as well?
```suggestion
    options?: { requeue?: boolean; revision?: string },
```

core/hierarchy/hierarchy-xwiki/src/components/componentsInit.ts
```diff
@@ -18,8 +18,8 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
 
-import { getRestSpacesApiUrl } from "../utils";
 import { name } from "@xwiki/cristal-hierarchy-api";
+import { getRestSpacesApiUrl } from "@xwiki/cristal-xwiki-utils";
```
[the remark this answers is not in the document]
pjeanjean: The previous one was a local import, so it was at the top, but it no longer needs to be there.

core/history/history-github/src/components/componentsInit.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Implementation of PageRevisionManager for the GitHub backend.
+ *
+ * @since 0.12
+ **/
+@injectable()
+class GitHubPageRevisionManager implements PageRevisionManager {
+  constructor(
+    @inject<CristalApp>("CristalApp") private cristalApp: CristalApp,
```
manuelleduc, line 37: We don't have a rule for that, but I tend to use `readonly` in addition to private for fields injected through the constructor.
```suggestion
    @inject<CristalApp>("CristalApp") private readonly cristalApp: CristalApp,
```

core/history/history-ui/src/vue/HistoryTab.vue
```diff
@@ -0,0 +1,117 @@
...
+<template>
+  <p>{{ t("history.extraTabs.header", { nRevisions: revisions.length }) }}</p>
+  <br />
+  <table>
+    <tr v-for="revision of revisions" :key="revision.version">
+      <td class="history-revision-cell">
+        <span class="history-revision">{{ revision.version }}</span>
+      </td>
+      <td>
+        <i18n-t keypath="history.extraTabs.entry" tag="span">
```
manuelleduc, line 64: Nice :)

core/history/history-ui/src/vue/HistoryTab.vue
```diff
@@ -0,0 +1,117 @@
...
+      .getContainer()
+      .get<PageRevisionManagerProvider>("PageRevisionManagerProvider")
+      .get()
+      .getRevisions(currentPage.value);
+  }
+});
+</script>
+
+<template>
+  <p>{{ t("history.extraTabs.header", { nRevisions: revisions.length }) }}</p>
```
manuelleduc, line 56: You should keep a `loading` state reference and display a loading message instead of the results during load.
Currently, it's displaying `This page has 0 versions stored.` while it's loading, which is not accurate.

core/history/history-xwiki/src/components/componentsInit.ts
```diff
@@ -0,0 +1,171 @@
...
+        Accept: "application/json",
+      };
+
+      if (authorization) {
+        headers.Authorization = authorization;
+      }
+      const response = await fetch(restApiUrl, { headers });
+      const jsonResponse = await response.json();
+      let user = jsonResponse.pageName;
+      jsonResponse.property.forEach(
```
manuelleduc, line 147: Just to be 100% sure, shouldn't it be plural (i.e., properties)?
```suggestion
      jsonResponse.properties.forEach(
```

pjeanjean: No, I checked
image

manuelleduc (approved):
Everything looks good. Thanks for putting up with me to finalize the last details.

PR #434, opened 2024-11-19
CRISTAL-235: Update navigation tree when creating/deleting/updating a page

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-235

# Changes

## Description

* This defines DocumentChange events in DocumentService, currently for page deletion and page update. Listeners can be registered and called anytime the DocumentService is notified of a change in the current Cristal instance.
* Both Shoelaces's and Vutify's Navigation Trees now implement DocumentChange listeners, and will update accordingly.

## Clarifications

Shoelace's Navigation Tree was completely refactored to avoid mutation observers. Instead, we now have a custom tree-item component exposing methods that can be called recursively from the tree component.

In the future, when we will implement moving documents, it can be supported by firing both a "delete" and an "update" DocumentChange.

# Screenshots & Video

N/A

# Executed Tests

The feature was tested manually, and the current integration tests were also run.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


skin/src/vue/c-sidebar.vue
```diff
@@ -172,6 +172,7 @@ function onClickOutsideMainSidebar() {
               .get<NavigationTreeSourceProvider>('NavigationTreeSourceProvider')
               .get()
           "
+          :document-service="documentService"
```
manuelleduc, line 175: same question here, wouldn't be easier to let  `XNavigationTree` resolve the document service internally?

core/document/document-default/src/defaultDocumentService.ts
```diff
@@ -100,10 +103,15 @@ function createStore(cristal: CristalApp): DocumentStoreDefinition {
 export class DefaultDocumentService implements DocumentService {
   private readonly refs: StateRefs;
   private readonly store: DocumentStore;
+  private documentChangeListeners: Map<
```
manuelleduc, line 106: Another option to directly initialize the field here, simplifying the constructor. 

```typescript
private readonly documentChangeListeners: Map<
    DocumentChange,
    Array<(page: PageData) => Promise<void>>
  > = new Map();
```

ds/shoelace/src/vue/x-navigation-tree-item.vue
```diff
@@ -0,0 +1,192 @@
...
+        const newItems = await props.treeSource.getChildNodes(props.node.id);
+        for (const newItem of newItems) {
+          if (newItem.id == nodes.value[i].id) {
+            nodes.value[i].label = newItem.label;
+            return;
+          }
+        }
+      } else {
+        parents.shift();
+        items.value![i].onDocumentUpdate(parents);
```
manuelleduc, line 136: missing await? asking to be sure

PR #450, opened 2024-11-27
CRISTAL-329: Creating a nested page fails to add a page with the right name

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-329

# Changes

## Description

* This PR fixes an issue with realtime editor that was fetching the wrong page on creation.
* Page creation was also improved to use SpaceReferences instead of manually crafting the new reference.
* While reviewing this issue, I detected a few edge cases on the navigation tree I did not catch before. It should now be more reliable.
* This also adds a small UI test that checks realtime editor opens properly, with the right data, on page creation.

## Clarifications

Fixing the issue with the realtime editor led to a new one: when the editor first saves a page that did not exist previously, the DocumentService is not updated with the currently edited page. This was not reproducible with the XWiki backend since the server always returns a document, even if the page doesn't exist yet.
So the editor now updates the current document after a save.

# Screenshots & Video

N/A

# Executed Tests

Reproduction steps were executed, and the test suite was executed locally.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/skin/src/vue/c-page-creation-menu.vue
+++ b/skin/src/vue/c-page-creation-menu.vue
@@ -55,7 +55,7 @@ function updateCurrentPage() {
 }
 
 function createPage() {
-  let newDocumentReference = new DocumentReference(
+  const newDocumentReference = new DocumentReference(
     name.value,
     locationReference!,
   );
--- a/skin/src/vue/c-page-creation-menu.vue
+++ b/skin/src/vue/c-page-creation-menu.vue
@@ -19,23 +19,35 @@ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
 -->
 <script setup lang="ts">
 import { CIcon } from "@xwiki/cristal-icons";
+import { DocumentReference, SpaceReference } from "@xwiki/cristal-model-api";
 import { defineProps, inject, ref } from "vue";
 import type { CristalApp, PageData } from "@xwiki/cristal-api";
+import type {
+  ModelReferenceSerializer,
+  ModelReferenceSerializerProvider,
+} from "@xwiki/cristal-model-reference-api";
 import type { NavigationTreeNode } from "@xwiki/cristal-navigation-tree-api";
 import type { Ref } from "vue";
 
 const cristal: CristalApp = inject<CristalApp>("cristal")!;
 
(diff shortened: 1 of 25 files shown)


skin/src/vue/c-page-creation-menu.vue
```diff
@@ -19,42 +19,51 @@ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
...
-  const pageResourceSeparator =
-    cristal.getWikiConfig().getType() == "XWiki" ? "." : "/";
-
-  if (location.value) {
-    newPage += location.value + pageResourceSeparator;
-  }
-  newPage += name.value;
-
-  cristal.setCurrentPage(newPage, "edit");
+  let newDocumentReference = new DocumentReference(
```
manuelleduc, line 58:
```suggestion
  const newDocumentReference = new DocumentReference(
```

web/e2e/main-page.spec.ts
```diff
@@ -171,6 +177,33 @@ configs.forEach(
       await expect(page.locator("#xwikicontent")).toContainText("Revision 2.1");
     });
 
+    test(`[${name}] has working editor on new page`, async ({ page, isMobile }) => {
+      await page.goto(localDefaultPage);
+
+      if (isMobile) {
+        const openSidebar = page.locator(".open-sidebar");
+        await openSidebar.nth(0).click();
+      }
```
manuelleduc, line 186: This should probably go to a Page Object.

manuelleduc (commented):
Looks good overall! I've added a few comments to address before merging.

web/e2e/pageObjects/Sidebar.ts
```diff
@@ -0,0 +1,64 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+import { expect } from "@playwright/test";
+import type { Locator, Page } from "@playwright/test";
+
+/**
+ * Page object to interact with the sidebar.
+ */
+export class SidebarPageObject {
```
manuelleduc, line 27: Missing since annotation. It can be convenient to have this information in the future.

manuelleduc (approved):
A small detail to fix and the PR is ready to merge.

PR #451, opened 2024-11-28
CRISTAL-300: CSS for the table displayer

tkrieck:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-300

# Changes

## Description

* Add basic styles for `<table>` elements

## Clarifications
* Additional styles for the `<a>`element were necessary to provide a good table reading experience;
* Because of these additional elements, some colors were imported to the basic `style.css`on each design system.

# Screenshots & Video

**Table on document content**
Screenshot 2024-11-28 at 10 26 38

**Attachments table**

Screenshot 2024-11-28 at 10 28 11

# Executed Tests

-

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/ds/shoelace/src/vue/css/style.css
+++ b/ds/shoelace/src/vue/css/style.css
@@ -83,7 +83,6 @@
   --cr-focus-ring: var(--sl-focus-ring-style) var(--sl-focus-ring-width) var(--sl-focus-ring-color);
   --cr-focus-ring-offset: 1px;
 
-
   /*
  * Color Primitives
  */
--- a/ds/shoelace/src/vue/css/style.css
+++ b/ds/shoelace/src/vue/css/style.css
@@ -33,6 +33,8 @@
   --cr-sizes-collapsed-main-sidebar-width: 32px; /*The width of the bar that gets shown containing the menu button to reopen the standard main main sidebar*/
   --cr-base-font-size: 16px;
   --cr-base-text-color: var(--cr-color-gray-950);
+  --cr-base-link-color: var(--cr-color-blue-600);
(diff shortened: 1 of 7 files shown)


ds/shoelace/src/vue/css/style.css
```diff
@@ -82,6 +84,244 @@
   --cr-focus-ring-offset: 1px;
 
 
+  /*
```
manuelleduc, line 87: bad formatting

ds/shoelace/src/vue/css/style.css
```diff
@@ -82,6 +84,244 @@
...
+ * Color Primitives
+ */
+
+  /* Gray */
+  --cr-color-gray-50: var(--sl-color-gray-50);
```
manuelleduc, line 92: This list of colors seems huge. Is everything used for this change? I'm wondering if it wouldn't be worth a separate improvement task.

manuelleduc (commented):
Overall looks good, except for some odd formatting.
I also added a comment regarding the introduction of th many color related CSS variables.

ds/shoelace/src/vue/css/style.css
tkrieck: Not all of them are used right now, but they are the basis for color usage in the future. In Shoelace I don't have to declare them individually because it has them incorporated, Vuetify on the other hand don't. Both DS are quite different in how they create themes, and I fear it is something that will get more and more inconvenient over time.

One option to keep them more aligned is to try to create a Cristal color theme and adapt both DS default themes to use those colors instead.

PR #481, opened 2024-12-06
CRISTAL-325, CRISTAL-343, CRISTAL-344, CRISTAL-346, CRISTAL-359

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-325
https://jira.xwiki.org/browse/CRISTAL-343
https://jira.xwiki.org/browse/CRISTAL-344
https://jira.xwiki.org/browse/CRISTAL-346
https://jira.xwiki.org/browse/CRISTAL-359

# Changes

## Description

On top of bugfixes:
* Introduce ModelReferenceHandler to have a way to manipulate ModelReferences (here, I added a method to create a DocumentReference from a given SpaceReference, depending on the backend)
* Support footers for x-dialogs on Vuetify (through a proper template slot)
* Delete `ds/shoelace/src/vue/x-textfield.vue` (it wasn't used anywhere and we had warning because the other textfield component had the same name, I assumed it was there by mistake)

## Clarifications

Sadly, the autofocus seems to be broken with dialogs on Shoelace (basically, it doesn't get triggered when the dialog is opened, but only on the initial component loading). I manage it right now with a pretty dirty fix and I don't think I can do better because of shadow roots, unless we use a lib such as [this one](https://github.com/edwinm/deep-query-selector-all) (or a custom function) but I'm not sure we want that.

# Screenshots & Video

N/A

# Executed Tests

The suite was run, bug fixed could not be reproduced with this PR applied.

# Expected merging strategy

* Prefers squash: No — This PR contains 5 commits that fix 5 different issues
* Backport on branches:
  * N/A


core/model/model-reference/model-reference-api/src/modelReferenceHandler.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+import type {
+  DocumentReference,
+  SpaceReference,
+} from "@xwiki/cristal-model-api";
+
+/**
```
manuelleduc, line 26: missing doc

core/model/model-reference/model-reference-api/src/modelReferenceHandler.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @since 0.13
+ */
+interface ModelReferenceHandler {
+  createDocumentReference(
```
manuelleduc, line 30: missing doc

core/model/model-reference/model-reference-xwiki/src/xWikiModelReferenceHandler.ts
```diff
@@ -0,0 +1,35 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+import { DocumentReference } from "@xwiki/cristal-model-api";
+import { injectable } from "inversify";
+import type { SpaceReference } from "@xwiki/cristal-model-api";
+import type { ModelReferenceHandler } from "@xwiki/cristal-model-reference-api";
+
+@injectable()
+export class XWikiModelReferenceHandler implements ModelReferenceHandler {
```
manuelleduc, line 27: given there is only two handlers, one for XWiki and one for the rest, wdyt of creating a default handler, and to use it as a fallback when the resolver don't find one with the request name?
That way we don't have to redefine the same new DocumentReference(name, space) all over again for each new backend.

ds/api/src/index.ts
```diff
@@ -47,6 +47,7 @@ type TextFieldProps = {
   name: string;
   label: string;
   required: boolean;
+  autofocus?: boolean;
```
manuelleduc, line 50: missing doc explaining what is this property for. I know it's missing for the rest, but we need to start somewhere.

ds/shoelace/src/components/shoelaceDesignSystemLoader.ts
```diff
@@ -37,11 +37,6 @@ export class ShoelaceDesignSystemLoader implements DesignSystemLoader {
     registerAsyncComponent(app, "XImg", () => import("../vue/x-img.vue"));
     registerAsyncComponent(app, "XRow", () => import("../vue/x-row.vue"));
     registerAsyncComponent(app, "XCol", () => import("../vue/x-col.vue"));
-    registerAsyncComponent(
-      app,
-      "XTextField",
-      () => import("../vue/x-textfield.vue"),
-    );
```
manuelleduc, line 44: Not sure to understand why this component is removed.

ds/shoelace/src/vue/x-dialog.vue
```diff
@@ -19,14 +19,39 @@ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
...
   width: string | number | undefined;
 }>();
 
+const dialog = useTemplateRef<SlDialog>("dialog");
+const slots = useSlots();
+
 function click() {
   open.value = true;
+
+  // TODO: autofocus in Shoelace's dialogs seems to be broken for the moment.
```
manuelleduc, line 36: Do you have the reference of the issue in shoelace bug tracker? Otherwise, we need to report it so that it get fixed.

ds/shoelace/src/vue/x-dialog.vue
```diff
@@ -36,14 +61,17 @@ const open = defineModel<boolean>();
...
-    v-if="open"
+    ref="dialog"
     :open="open"
     :label="title"
     class="dialog-overview"
     @sl-show="open = true"
     @sl-hide="open = false"
   >
     <slot name="default" />
+    <div v-if="slots.footer" slot="footer">
```
manuelleduc, line 72: This piece deserved a comment explaining why we need to do what we do.
I think I understand it, but I feel like I'll need to think hard again if I have to understand it again, without a proper comment.

manuelleduc (commented):
Looks good overall.
Please check that all new components have proper documentation :)

ds/shoelace/src/components/shoelaceDesignSystemLoader.ts
pjeanjean: I mentioned it in the changes: 
> it wasn't used anywhere and we had warning because the other textfield component had the same name, I assumed it was there by mistake

ds/shoelace/src/vue/x-dialog.vue
pjeanjean: While trying to make a reproducible example to open an issue, I ended up noticing that it was actually a compatibility issue with Vue3, and it could be fixed using only Vue3. I will document this because it's likely similar things could happen with Shoelace component.

In this case, the issue was that the `autofocus` prop was passed to the children nodes but not rendered in the DOM, so even though the `input` tag had it, its ancestor `sl-input` did not. This seems to be the expected behavior on Vue3 when using `v-bind`, and the way to fix it is to add the `.attr` modifier that I cannot find mentioned in Vue's doc at all...

ds/shoelace/src/components/shoelaceDesignSystemLoader.ts
manuelleduc: Indeed, good point. Thanks.
This change should be part of a `[Misc]` commit imo.

core/model/model-reference/model-reference-api/src/defaultModelReferenceHandlerProvider.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+@injectable()
+class DefaultModelReferenceHandlerProvider
+  implements ModelReferenceHandlerProvider
+{
+  constructor(
+    @inject<CristalApp>("CristalApp") private cristalApp: CristalApp,
+  ) {}
+
+  get(type?: string): ModelReferenceHandler | undefined {
```
manuelleduc, line 37: A comment explaining the fallback strategy wouldn't hurt here.

core/model/model-reference/model-reference-api/src/defaultModelReferenceHandlerProvider.ts
```diff
@@ -0,0 +1,48 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+import { ModelReferenceHandler } from "./modelReferenceHandler";
+import { ModelReferenceHandlerProvider } from "./modelReferenceHandlerProvider";
+import { inject, injectable } from "inversify";
+import type { CristalApp } from "@xwiki/cristal-api";
+
+/**
+ * @since 0.13
```
manuelleduc, line 27: As a general rule, just a `since` annotation is not a good sign.
Interfaces should explain their role.
Components should explain their specificities in regard to other implementations.

core/model/model-reference/model-reference-filesystem/src/componentInit.ts
```diff
@@ -20,11 +20,11 @@
 
 import { FileSystemModelReferenceParser } from "./filesystemModelReferenceParser";
 import { FileSystemModelReferenceSerializer } from "./filesystemModelReferenceSerializer";
-import {
+import type {
```
manuelleduc, line 23: shouldn't this be part of a separate `Misc` commit?

ds/api/src/index.ts
```diff
@@ -43,10 +43,16 @@ type TabProps = {
...
 
+/**
+ * Props of the text-field component.
+ * @since 0.9
+ */
 type TextFieldProps = {
   name: string;
   label: string;
   required: boolean;
+  /** whether the field should be focused on load or not (default: false) */
```
manuelleduc, line 54: Individual fields can have specific `since` annotations, to override the more general `since` of the type itself.

core/model/model-reference/model-reference-api/src/defaultModelReferenceHandlerProvider.ts
```diff
@@ -0,0 +1,55 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+import { ModelReferenceHandler } from "./modelReferenceHandler";
+import { ModelReferenceHandlerProvider } from "./modelReferenceHandlerProvider";
+import { inject, injectable } from "inversify";
+import type { CristalApp } from "@xwiki/cristal-api";
+
+/**
+ * Default implementation for ModelReferenceHandlerProvider. Will provide an
```
manuelleduc, line 27: Even though we don't yet generate document, a `{@link ModelReferenceHandlerProvider}` syntax exists, which is nice to:
1. make it clear this part of the text is a link to an existing entity
2. make the generated doc nicer once we start generating it.

PR #504, opened 2024-12-10
CRISTAL-364: Breadcrumb links not working with Shoelace

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-364

# Changes

## Description

The previous Breadcrumb implementation was using direct XWiki links, obtained from XWiki's API, and relied on a "click listener" set on the links. However, after the refactoring done for [CRISTAL-274](https://jira.xwiki.org/browse/CRISTAL-274), the listener seems to no longer work properly with Shoelace components (that use a lot of shadow DOMs). As such, this implementation drops the listener and makes proper local Cristal links instead, which should be more stable.

## Clarifications

While using the ModelReference API for this PR, I noticed a few edge cases that were not properly handled (my test instance has a few documents with special characters, and they were not properly escaped). So I fixed them in both the parser and serializer, and also added a unit test for this.

# Screenshots & Video

N/A

# Executed Tests

The test suite was executed locally.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/core/hierarchy/hierarchy-xwiki/package.json
+++ b/core/hierarchy/hierarchy-xwiki/package.json
@@ -27,6 +27,9 @@
     "@xwiki/cristal-authentication-api": "workspace:*",
     "@xwiki/cristal-backend-api": "workspace:*",
     "@xwiki/cristal-hierarchy-api": "workspace:*",
+    "@xwiki/cristal-model-api": "workspace:*",
+    "@xwiki/cristal-model-reference-api": "workspace:*",
+    "@xwiki/cristal-model-remote-url-api": "workspace:*",
     "@xwiki/cristal-xwiki-utils": "workspace:*",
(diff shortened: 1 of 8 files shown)


manuelleduc (commented):
Looks good, thanks for fixing this.
Can I let you move everything unrelated to the click handling (i.e., the part related to special chars handling) in a separate PR?
1. create a bug issue for this part
2. have a PR with the new issue id

manuelleduc (commented):
Also, to anticipate further regression, wdyt of adding an e2e test with a page navigation using the breadcrumb?

pjeanjean:
Added UI test and moved special characters handling to https://github.com/xwiki-contrib/cristal/pull/507

manuelleduc (approved):

PR #522, opened 2024-12-17
CRISTAL-371: Shows details about the last edition on pages

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-371

# Changes

## Description

This PR adds the properties `lastAuthor` and `lastModificationDate` to `PageData`. When set, these properties are used to display the last edition details in `c-article`.
They are set for two backends: FileSystem and XWiki.

## Clarifications

I am making use of the pluralization feature of `vue-i18n` to choose between a localization string with or without user name. I don't think we have done something similar in the past, so I'm open to feedback on this.

# Screenshots & Video

image

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


api/src/api/PageData.ts
```diff
@@ -32,6 +32,16 @@ export interface PageData {
...
+  /**
+   * Date of the last modification of this page.
+   * @since 0.13
+   */
+  lastModificationDate: Date | undefined;
+  /**
+   * Name of the last user to edit this page.
+   * @since 0.13
+   */
+  lastAuthor: string | undefined;
```
manuelleduc, line 44: Wouldn't it be better to use `UserDetails` from `@xwiki/cristal-authentication-api`?

core/backends/backend-nextcloud/src/nextcloudStorage.ts
```diff
@@ -71,13 +71,50 @@ export class NextcloudStorage extends AbstractStorage {
...
       if (response.status >= 200 && response.status < 300) {
+        let lastAuthor: string | undefined;
+        let lastModificationDate: Date | undefined;
+        const propResponse = await fetch(
+          `${this.getWikiConfig().baseRestURL}/${USERNAME}/.cristal/${page}/page.json`,
+          {
+            body: `<?xml version="1.0" encoding="UTF-8"?>
+              <d:propfind xmlns:d="DAV:">
+                <d:prop xmlns:oc="http://owncloud.org/ns">
+                  <d:getlastmodified />
```
manuelleduc, line 82: So you know if it is possible to access the content though PROPFIND, to avoid having to chain two HTTP requests to obtain the result?

skin/langs/translation-en.json
```diff
@@ -6,6 +6,7 @@
   "page.creation.menu.field.name": "Name",
   "page.creation.menu.submit": "Create",
   "page.creation.menu.title": "New Page",
+  "page.edited.details": "Edited on {date} | {user} edited on {date}",
```
manuelleduc, line 9: I thought the `|` was for pluralization. Does it also work as a fallback when `user` is not defined?

core/backends/backend-nextcloud/src/nextcloudStorage.ts
manuelleduc: ok, thanks for checking

core/backends/backend-nextcloud/src/nextcloudStorage.ts
```diff
@@ -71,13 +72,52 @@ export class NextcloudStorage extends AbstractStorage {
       );
 
       if (response.status >= 200 && response.status < 300) {
+        let lastAuthor: UserDetails | undefined;
```
manuelleduc, line 75: WDYT of moving the last author/date logic in private method, to prevent `getPageContent` to become too large?

core/backends/backend-nextcloud/src/nextcloudStorage.ts
```diff
@@ -127,6 +93,45 @@ export class NextcloudStorage extends AbstractStorage {
     }
   }
 
+  private async getLastEditDetails(
+    page: string,
+  ): Promise<[Date | undefined, UserDetails | undefined]> {
```
manuelleduc, line 98: I think we should always favor named return values instead of indexed ones
```suggestion
  ): Promise<{date?: Date, userDetails?: UserDetails}> {
```

skin/src/vue/c-article.vue
```diff
@@ -72,7 +74,27 @@ watch(
...
         <div class="info-wrapper">
           <span class="doc-author">
-            <!-- Add last edition information (CRISTAL-371). -->
+            <i18n-t
+              v-if="currentPage?.lastAuthor"
+              keypath="page.edited.details.user"
+              tag="span"
+            >
+              <template #date>
+                <date :date="currentPage?.lastModificationDate!" />
```
manuelleduc, line 83: Shouldn't we just hide the whole element for backend that are not able to compute a last modification date?
The alternative is to make this information mandatory, as I can't imagine a backend that wouldn't be able to provide such information.

PR #530, opened 2024-12-20
CRISTAL-382: No clear message when the current user is not allowed to view a page

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-382

# Changes

## Description

* Add Provider for AlertsService, in order to inject it even before the initialization of pinia.
* Detect duplicated alerts and show the number of duplications.

## Clarifications

This includes a Misc commit that removes an unnecessary page load.

# Screenshots & Video

image

# Executed Tests

The test suite was run fully.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * No — There is also a Misc commit

### DIFF
--- a/core/alerts/alerts-api/src/index.ts
+++ b/core/alerts/alerts-api/src/index.ts
@@ -31,6 +31,11 @@ interface Alert {
   type: Type;
   message: string;
   actions?: Action[];
+  /**
+   * Number of times this specific alert has already been created.
+   * @since 0.13
+   */
+  duplicatesCount: number;
 }
 
 /**
@@ -80,4 +85,17 @@ interface AlertsService {
   dismiss(id: number): void;
 }
 
-export type { Action, Alert, AlertsService, Type };
+/**
+ * Helper to resolve an instance of {@link AlertsService} lazily.
+ *
+ * @since 0.13
+ */
+interface AlertsServiceProvider {
+  /**
+   * Get an instance of {@link AlertsService}.
+   * @returns the instance wanted
+   */
+  get(): AlertsService;
+}
+
+export type { Action, Alert, AlertsService, AlertsServiceProvider, Type };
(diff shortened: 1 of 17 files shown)


core/alerts/alerts-api/src/index.ts
```diff
@@ -31,6 +31,11 @@ interface Alert {
   type: Type;
   message: string;
   actions?: Action[];
+  /**
+   * Number of times this specific alert has already been created.
+   * @since 0.13
+   */
+  nDuplicates: number;
```
manuelleduc, line 38: WDYT of?
```suggestion
  duplicatesCount: number;
```

core/alerts/alerts-default/src/defaultAlertsService.ts
```diff
@@ -62,12 +62,20 @@ const alertsStore: AlertsStoreDefinition = defineStore<
...
-        id: this.id++,
-        type: type,
-        message: message,
-        actions: actions,
-      });
+      const foundAlert = this.alerts.find(
+        (a) => a.type == type && a.message == message && a.actions == actions,
+      );
+      if (foundAlert) {
+        foundAlert.nDuplicates++;
```
manuelleduc, line 69: I don't remember the details. Does the alters stays in memory forever, are they removed from the list after a defined amont or time (and on user action)?
If that's the case, it might be better to compute the number of duplicated on display. Instead of keeping a stateful internal variable.

pjeanjean: They do not stay in memory forever, right now they are removed when `dismiss()` is called, which happens when clicking the "X" button. I think it's easier to handle in the service, this keeps the UI part cleaner with a simple iteration on the list without any filtering.

core/alerts/alerts-api/src/index.ts
pjeanjean: Agreed, that's better.

core/alerts/alerts-default/src/defaultAlertsService.ts
manuelleduc: So, to be sure I understand fully the counter falls down to 0 only when all the "same" messages are dismissed?

pjeanjean: The "same" messages are just not stored, there is only one (with the counter) and it will be the only one until it is dismissed. Then, if it comes back, a new one is created with a 0 counter.

PR #540, opened 2024-12-23
CRISTAL-257: The edit button is presented even if the current user does not have edit rights

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-257

# Changes

## Description

* Store whether or not the current user can edit the returned page.
* Hide the page actions if not.
* Show error on failed save.

## Clarifications

The changes in this PR are only related to the XWiki backend.

# Screenshots & Video

image

# Executed Tests

The test suite was run fully, and the new feature was tested with multiple rights combinations.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


core/backends/backend-xwiki/src/xwikiStorage.ts
```diff
@@ -315,6 +316,14 @@ export class XWikiStorage extends AbstractStorage {
       // TODO: the syntax provided by the save is ignored and the content is always saved as markdown.
       body: JSON.stringify({ content, title, syntax: "markdown/1.2" }),
     });
+    if (!response.ok) {
+      const errorMessage = await response.text();
+      // TODO: Fix CRISTAL-383
```
manuelleduc, line 321: Maybe explain what's the issue with CRISTAL-383

core/backends/backend-xwiki/src/xwikiStorage.ts
```diff
@@ -315,6 +316,14 @@ export class XWikiStorage extends AbstractStorage {
...
+      this.alertsServiceProvider
+        .get()
+        .error(`Could not save page ${page}. Reason: ${errorMessage}`);
+      throw new Error(errorMessage);
```
manuelleduc, line 325: Do we really need to throw an error if it is already handled by the alert service provider?

skin/src/vue/c-content.vue
```diff
@@ -113,20 +113,25 @@ onUpdated(() => {
       <div class="doc-page-actions">
         <router-link
           :to="
-            currentPageRevision
+            currentPageRevision || !currentPage?.canEdit
```
manuelleduc, line 116: Wouldn't it be better to simply hide the edit button when the user does not have the right to edit?

core/backends/backend-xwiki/src/xwikiStorage.ts
pjeanjean: Actually, yes, we need it to update the editor status (that shows "Unsaved" only if it catches an error).

manuelleduc: ok, maybe add a comment to explain that, so that nobody removes it by accident.

pjeanjean: Good point.

PR #569, opened 2025-01-13
CRISTAL-414: Color and label issue for link in messages

tkrieck:
* Fixes the color of links inside the message component of Vuetify
* Small correction on a message localization for EN language

# Jira URL

https://jira.xwiki.org/browse/CRISTAL-414

# Changes

## Description

* Fix for link colors inside the message component of Vuetify

## Clarifications

* Also fixes a small translating issue in English

# Screenshots & Video

Before: 
image

After:

Screenshot 2025-01-13 at 09 30 42

# Executed Tests

-

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/ds/vuetify/src/vue/x-alert.vue
+++ b/ds/vuetify/src/vue/x-alert.vue
@@ -37,3 +37,11 @@ defineProps<AlertProps>();
     <slot />
   </v-alert>
 </template>
+
+<style scoped>
+:deep(.v-alert__content a.router-link-active),
+:deep(.v-alert__content a.router-link-active:visited) {
+  color: inherit;
+  text-decoration-color: inherit;
+}
+</style>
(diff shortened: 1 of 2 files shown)


manuelleduc (approved):
The fix looks good, thanks!
I just have a comment regarding the description of the issue. I think a sentence describing the issue would be better, something along "Color and label issue for link in x-alert components".
WDYT?

tkrieck:
> The fix looks good, thanks! I just have a comment regarding the description of the issue. I think a sentence describing the issue would be better, something along "Color and label issue for link in x-alert components". WDYT?

I mostly agree, just the name of the affected component can be hard to figure out before opening the issue. But otherwise, I'm on board. You need me to adjust the issue or leave this one as is?

PR #580, opened 2025-01-16
CRISTAL-408: We don't check if a page already exists on page creation

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-408

# Changes

## Description

This adds an alert in the page creation dialog if the page already exists, with an action to edit it instead.

## Clarifications

N/A

# Screenshots & Video

image

# Executed Tests

The UI tests were run locally.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/skin/langs/translation-en.json
+++ b/skin/langs/translation-en.json
@@ -1,8 +1,9 @@
 {
   "third": "Third",
   "page.actions.create.label": "Create a new document",
-  "page.creation.menu.alert.action.edit": "Edit",
-  "page.creation.menu.alert.content": "The page {pageName} already exists, either use another name or edit the existing page.",
+  "page.creation.menu.alert.content": "The page {pageName} already exists, use another name.",
+  "page.creation.menu.alert.content.edit": "The page {pageName} already exists, either use another name or {link}.",
+  "page.creation.menu.alert.content.edit.link": "edit the existing page",
   "page.creation.menu.button": "New Page",
   "page.creation.menu.field.location": "Parent location",
   "page.creation.menu.field.name": "Name",
--- a/skin/langs/translation-en.json
+++ b/skin/langs/translation-en.json
@@ -1,6 +1,8 @@
 {
   "third": "Third",
   "page.actions.create.label": "Create a new document",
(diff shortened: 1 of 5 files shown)


skin/src/vue/c-page-creation-menu.vue
```diff
@@ -141,6 +176,13 @@ function createPage() {
 </template>
 
 <style scoped>
+#new-page-alerts-wrapper {
```
manuelleduc, line 179: I think using classes would be better than ids here, unless there is a reason?

pjeanjean:
Updated screenshots:
image
image

skin/langs/translation-en.json
```diff
@@ -1,6 +1,9 @@
 {
   "third": "Third",
   "page.actions.create.label": "Create a new document",
+  "page.creation.menu.alert.content": "The page {pageName} already exists, use another name.",
```
manuelleduc, line 4: Maybe just me but I feel like explaining the page already exists is enough.
```suggestion
  "page.creation.menu.alert.content": "The page {pageName} already exists.",
```

skin/langs/translation-fr.json
```diff
@@ -1,6 +1,9 @@
 {
     "third": "Troisième",
     "page.actions.create.label": "Créer un nouveau document",
+    "page.creation.menu.alert.content": "La page {pageName} existe déjà, choisissez un autre nom.",
+    "page.creation.menu.alert.content.edit": "La page {pageName} existe déjà, choisissez un autre nom ou {link}.",
+    "page.creation.menu.alert.content.edit.link": "éditez la page existante",
```
manuelleduc, line 6: This goes against https://dev.xwiki.org/xwiki/bin/view/Community/L10N/Conventions/#HRulesfortranslatingcompositeandlongmessages

> Do not split sentences into smaller translated fragments.

But I don't know better for "nested" message subpart that are expected to be wrapped with html dynamically.

manuelleduc (approved):
Looks good, just a small comment on translation string value.

PR #653, opened 2025-02-05
CRISTAL-462: page title have a slightly different style on edit mode

tkrieck:
* Adjusted styles for the page title

# Jira URL

https://jira.xwiki.org/browse/CRISTAL-462

# Changes

## Description

Slight variation in style of the doc title between edit and view mode.

## Clarifications

* There was a style that was being inherited by the input in edit mode that made the component differ a little bit from the view mode component, this was adjusted by adding a property directly to the class `doc-title`

# Screenshots & Video

Before: 

Screenshot 2025-02-05 at 14 26 39

After:

Screenshot 2025-02-05 at 14 26 11

# Executed Tests

 -

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/skin/src/vue/c-content.vue
+++ b/skin/src/vue/c-content.vue
@@ -160,6 +160,7 @@ onUpdated(() => {
   font-size: var(--cr-font-size-2x-large);
   line-height: var(--cr-font-size-2x-large);
   padding-block-start: var(--cr-spacing-small);
+  font-weight: var(--cr-font-weight-bold);
 }
 
 .doc-page-actions {


skin/src/vue/c-content.vue
```diff
@@ -160,6 +160,7 @@ onUpdated(() => {
   font-size: var(--cr-font-size-2x-large);
   line-height: var(--cr-font-size-2x-large);
   padding-block-start: var(--cr-spacing-small);
+  font-weight: var(--cr-font-weight-bold);
```
manuelleduc, line 163: I think this is leading to more inconsistencies in the future as the style is duplicated in c-content and c-edit-tiptap.

I'd suggest to add the code below in c-article, and to remove the corresponding properties in the `.doc-title` selector of c-content and c-edit-tiptap.

WDYT?

```css
:deep(.doc-title) {
  margin: 0;
  font-size: var(--cr-font-size-2x-large);
  line-height: var(--cr-font-size-2x-large);
  font-weight: var(--cr-font-weight-bold);
  padding-block-start: var(--cr-spacing-small);
}
```

skin/src/vue/c-content.vue
```diff
@@ -154,14 +154,6 @@ onUpdated(() => {
   height: auto;
 }
 
-.doc-title {
-  grid-area: doc-title;
```
manuelleduc, line 158: This property is not part of the definition on c-article. Is it expected?
Same question for `padding-block-start` below.

tkrieck: Good catch, thank you. Indeed it was a mistake, added them back in ea038d8d1e1ebd03beb1dce373c6f8068d0cd200

PR #678, opened 2025-02-19
CRISTAL-465: Make Electron app work properly again

ClementEXWiki:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-465

# Changes

## Description

* Fix a timing issue causing Electron to start with a blank window

## Clarifications

The problem seems to be caused by a recent Electron update

# Screenshots & Video

Before:

image

After:

image

# Executed Tests

Manual testing

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/lib/src/components/cristalAppLoader.ts
+++ b/lib/src/components/cristalAppLoader.ts
@@ -161,7 +161,8 @@ class CristalAppLoader extends CristalLoader {
       const localConfigName = window.localStorage.getItem("currentApp");
       if (
         localConfigName != null &&
-        this.cristal.getAvailableConfigurations().has(localConfigName)
+        (!this.cristal ||
+          this.cristal.getAvailableConfigurations().has(localConfigName))
       ) {
         return localConfigName;
       } else {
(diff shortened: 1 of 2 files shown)


lib/src/components/cristalAppLoader.ts
```diff
@@ -161,7 +161,8 @@ class CristalAppLoader extends CristalLoader {
       const localConfigName = window.localStorage.getItem("currentApp");
       if (
         localConfigName != null &&
-        this.cristal.getAvailableConfigurations().has(localConfigName)
+        (!this.cristal ||
```
manuelleduc, line 164: Thanks! I confirm this is fixing the issue when trying locally, both at startup and when switching configuration.
One last thing before merging, this change is worth an explanation of which is additional check if required.

ClementEXWiki: Yes, basically `this.cristal` may not be defined when the class is instanciated, hence the check.
I'm not 100% sure this is a correct fix, as while it fixes the problem I don't know if the check is mandatory in some scenarios. I'm not familiar enough with the codebase to be sure about that.

manuelleduc:
Manually merged with https://github.com/xwiki-contrib/cristal/commit/8133dc7d70201b688a7a51262d9cb28d3a0cb5ae

PR #681, opened 2025-02-19
[Misc] Pin Node.js version when using mise-en-place

ClementEXWiki:
# Jira URL

N/A

# Changes

## Description

* Pin Node.js version when using [mise-en-place](https://mise.jdx.dev/)

## Clarifications

* Mise is a tool suite enabling notably the management of multiple Node.js versions. This PR allows it to automatically select and download the correct Node.js version for the project.

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

cc @manuelleduc


mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
+node = "22.13.1"
```
manuelleduc, line 2: I'm not a fan of seeing the node version duplicated between this file and the root `package.json`
Would it be possible to point this config file to the package.json definition?

ClementEXWiki:
Just checked and mise-en-place does support the `.node-version` file. This could be interesting as it's a more universal approach.

PR #683, opened 2025-02-19
[Misc] Use re-export statements instead of import -> export

ClementEXWiki:
# Jira URL

N/A

# Changes

## Description

* Replace `import` followed by `export` with re-`export` statements

## Clarifications

* Instead of `import`ing a module and then `export`ing it directly, it's preferred to use a re-`export` statement:

```typescript
// Instead of:
import { obj } from './module'
export { obj }

// Write this:
export { obj } from './module'
```

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/api/src/index.ts
+++ b/api/src/index.ts
@@ -18,46 +18,22 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
 
-import { registerAsyncComponent } from "./api/designSystemLoader.js";
-import { DefaultPageData } from "./components/DefaultPageData.js";
-import { JSONLDDocument } from "./components/JSONLDDocument.js";
-import ComponentInit from "./components/componentsInit.js";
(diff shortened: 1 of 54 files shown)


ClementEXWiki:
cc @manuelleduc

manuelleduc:
Thanks for working on this.
I understand it makes thinks more compact, but this is not our current coding style.
What I suggest is to:
1. write a proposal on https://forum.xwiki.org/c/cristal/18 to propose a new best practice
2. ideally find a configuration for eslint to enforce this coding style, avoiding future inconsistencies

ClementEXWiki:
Done in https://forum.xwiki.org/t/coding-style-use-re-exports-instead-of-import-export/16504

PR #685, opened 2025-02-20
[Misc] Code style: various improvements

ClementEXWiki:
# Jira URL

N/A

# Changes

## Description

* Code style improvements

## Clarifications

N/A

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/api/src/components/defaultWikiConfig.ts
+++ b/api/src/components/defaultWikiConfig.ts
@@ -89,25 +89,24 @@ export class DefaultWikiConfig implements WikiConfig {
     this.offline = offline;
   }
 
-  setConfigFromObject(configObject: Record<string, unknown>): void {
-    // TODO: validate object shape with a library
-
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.name = configObject.name as any;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.baseURL = configObject.baseURL as any;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.baseRestURL = configObject.baseRestURL as any;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.realtimeURL = configObject.realtimeURL as any;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.homePage = configObject.homePage as any;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.serverRendering = configObject.serverRendering as any;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.offline = configObject.offline as any;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.designSystem = configObject.designSystem as any;
+  setConfigFromObject(configObject: {
+    name: string;
+    baseURL: string;
+    baseRestURL: string;
+    realtimeURL: string;
+    homePage: string;
+    serverRendering: boolean;
+    offline: boolean;
+    designSystem: string;
+  }): void {
+    this.name = configObject.name;
+    this.baseURL = configObject.baseURL;
+    this.baseRestURL = configObject.baseRestURL;
+    this.realtimeURL = configObject.realtimeURL;
+    this.homePage = configObject.homePage;
+    this.serverRendering = configObject.serverRendering;
+    this.offline = configObject.offline;
+    this.designSystem = configObject.designSystem;
   }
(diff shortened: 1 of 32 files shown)


core/backends/backend-github/src/githubStorage.ts
```diff
@@ -61,7 +61,7 @@ export class GitHubStorage extends AbstractStorage {
     return `${this.wikiConfig.baseRestURL}${directory}${image}`;
   }
 
-  hashCode = function (str: string): string {
+  hashCode = (str: string): string => {
```
manuelleduc, line 64: This change is unrelated to the topic of this PR.
Also, we currently don't have a rule regarding the use of `function` vs `=>` for anonymous functions.

core/backends/backend-xwiki/src/xwikiStorage.ts
```diff
@@ -272,7 +272,7 @@ export class XWikiStorage extends AbstractStorage {
     try {
       // TODO get rid of any
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const fieldMapping: any = document.get("xwikiMapping")[fieldName];
+      const fieldMapping = (document.get("xwikiMapping") as any)[fieldName];
```
manuelleduc, line 275: I'm not sure to understand why this is an improvement. In both cases there is a type coercion leading to `any`

api/src/components/defaultWikiConfig.ts
```diff
@@ -89,17 +89,25 @@ export class DefaultWikiConfig implements WikiConfig {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.realtimeURL = configObject.realtimeURL as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.homePage = configObject.homePage as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.serverRendering = configObject.serverRendering as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.offline = configObject.offline as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.designSystem = configObject.designSystem as any;
```
manuelleduc, line 110: Isn't it a more verbose way to achieve the same thing as before?

core/hierarchy/hierarchy-xwiki/src/components/componentsInit.ts
```diff
@@ -90,32 +90,30 @@ class XWikiPageHierarchyResolver implements PageHierarchyResolver {
...
-                    )! as DocumentReference,
-                  ),
-                },
-              }).href,
-            });
-          }
-        },
-      );
+
+      for (const hierarchyItem of jsonResponse.hierarchy.items) {
```
manuelleduc, line 94: This change is underrated to the topic of this PR.
Also, we don't have a rule for the use of `forEach` vs `for...of`

manuelleduc (changes requested):
Many changes in this PR are not related to the removal of the use of `any`.
I think it would be easier to discuss them in separate PRs.

ClementEXWiki:
Sorry, the PR's title was based on the first commit, but there are multiple ones.
Fixed the title

core/backends/backend-xwiki/src/xwikiStorage.ts
ClementEXWiki: The `any` marker is moved from the whole expression to just the `document.get` portion
It is indeed to very importrant

core/hierarchy/hierarchy-xwiki/src/components/componentsInit.ts
ClementEXWiki: There may not be a rule for this but `for..of` is usually faster than `.forEach`

api/src/components/defaultWikiConfig.ts
ClementEXWiki: Not really, as this explicitly states that `configObject` is a `Record` instead of being an `any`.
But given we need to have lots of properties, it would make more sense to have a complete type to this argument ; I've just changed it in the PR

core/backends/backend-github/src/githubStorage.ts
ClementEXWiki: Arrow functions are the more modern equivalent ; they also avoid creating their own `this` scope which can be a footgun when using the `function () {}` syntax.

PR #688, opened 2025-02-20
[Misc] Code style: get rid of 'any' types

ClementEXWiki:
# Jira URL

N/A

# Changes

## Description

* Remove usages of the `any` type

## Clarifications

N/A

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/api/src/api/WikiConfig.ts
+++ b/api/src/api/WikiConfig.ts
@@ -20,7 +20,18 @@
 
 import type { Storage } from "./storage";
 
-export interface WikiConfig {
+type ConfigObject = {
+  name: string;
+  baseURL: string;
+  baseRestURL: string;
+  realtimeURL: string;
+  homePage: string;
+  serverRendering: boolean;
+  offline: boolean;
+  designSystem: string;
+};
+
+interface WikiConfig {
   name: string;
   // The base url of the backend endpoint
   baseURL: string;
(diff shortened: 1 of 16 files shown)


api/src/api/WikiConfig.ts
```diff
@@ -20,7 +20,18 @@
 
 import type { Storage } from "./storage";
 
-export interface WikiConfig {
+type ConfigObject = {
```
manuelleduc, line 23: If you export something please define which version it was introduced in with the `@since` annotation.
An explanation of the type in the tsdoc is also preferable.

api/src/components/DefaultPageData.ts
```diff
@@ -73,17 +71,26 @@ export class DefaultPageData implements PageData {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.syntax = object.syntax as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.html = object.html as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.css = object.css as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.js = object.js as any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    this.version = object.version as any;
```
manuelleduc, line 94: I'm not sure how this is better than simply definition the type of the `object` parameter as `any`

PR #690, opened 2025-02-20
[Misc] Code style: use for loops instead of .forEach()

ClementEXWiki:
# Jira URL

N/A

# Changes

## Description

* Use `for`..`of` loops instead of `.forEach()`

## Clarifications

* `for`..`of` loops are faster at runtime

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


manuelleduc:
Thanks!
A is important to also add the corresponding eslint rules (e.g., https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md) when they exist.
The same is true for the other PRs of the same kind.

PR #796, opened 2025-03-31
CRISTAL-496: Allow the file system backend root location to be configurable

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-496

# Changes

## Description

This PR adds support for the `storageRoot` configuration key to the FileSystem backend.

## Clarifications

The `storageRoot` value needs to be made available to the main process of electron. In order to do this, a state package was created at `electron/state` to set and get this information.
BrowserApi and its Electron implementation were slightly refactored to handle that.

# Screenshots & Video

N/A

# Executed Tests

The test suite was run locally.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/core/browser/browser-default/src/components/browser-api-default.ts
+++ b/core/browser/browser-default/src/components/browser-api-default.ts
@@ -49,4 +49,6 @@ export class BrowserApiDefault implements BrowserApi {
       window.removeEventListener("beforeunload", listener);
     });
   }
+
+  setLocation(): void {}
 }
--- a/core/browser/browser-default/src/components/browser-api-default.ts
+++ b/core/browser/browser-default/src/components/browser-api-default.ts
@@ -50,5 +50,9 @@ export class BrowserApiDefault implements BrowserApi {
     });
   }
 
-  setLocation(): void {}
+  setLocation(): void {
+    // At this point, we need to change the current URL on configuration
+    // switching, but this operation is already handled during app loading for
+    // the initial configuration (which is when this method is called).
+  }
 }
--- a/electron/state/src/index.ts
+++ b/electron/state/src/index.ts
@@ -22,10 +22,20 @@ const storedConfig: {
   storageRoot?: string;
 } = {};
 
+/**
+ * Set the storage root location for the current configuration.
+ * @param storageRoot - the new storage root location
+ * @since 0.16
+ */
 function setStorageRoot(storageRoot?: string): void {
   storedConfig.storageRoot = storageRoot;
 }
 
+/**
+ * Get the storage root location for the current configuration.
+ * @returns the current storage root location
+ * @since 0.16
+ */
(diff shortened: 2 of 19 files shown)


core/browser/browser-default/src/components/browser-api-default.ts
```diff
@@ -49,4 +49,6 @@ export class BrowserApiDefault implements BrowserApi {
       window.removeEventListener("beforeunload", listener);
     });
   }
+
+  setLocation(): void {}
```
manuelleduc, line 53: A comment explaining why the method is empty could be interesting.

electron/state/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+function setStorageRoot(storageRoot?: string): void {
+  storedConfig.storageRoot = storageRoot;
+}
+
+function getStorageRoot(): string | undefined {
+  return storedConfig.storageRoot;
+}
+
+export { getStorageRoot, setStorageRoot };
```
manuelleduc, line 33: missing since and documentation

electron/storage/src/components/FileSystemConfig.ts
```diff
@@ -43,4 +46,11 @@ export class FileSystemConfig extends DefaultWikiConfig {
   override getType(): string {
     return "FileSystem";
   }
+
+  override initialize(): void {
+    super.initialize();
+    // The initialization for the root directory needs to be done after a
+    // configuration change, but before any storage operation.
+    fileSystemStorage.initRootDirectory();
```
manuelleduc, line 54: Just to confirm, as it impacts directories initialization. Did you manually test several configurations such as the one below?
- pointing to an empty `~/.cristal`
- pointing to an non-empty `~/.cristal`
- pointing to an empty `~/.somethingelse`
- pointing to an non-empty `~/.somethingelse`

electron/storage/src/electron/main/index.ts
```diff
@@ -32,9 +33,12 @@ import { basename, dirname, join, relative } from "node:path";
 const HOME_PATH = ".cristal";
 const HOME_PATH_FULL = join(app.getPath("home"), HOME_PATH);
 
+function getHomePathFull(): string {
+  return getStorageRoot() ?? HOME_PATH_FULL;
```
manuelleduc, line 37: An explanation of the fallback could be interesting here. Just to clarify that, we fall back to the default path when no path is explicitly provided in the configuration.

electron/storage/src/components/FileSystemConfig.ts
pjeanjean: Yes, I also tried these configurations both on start (with the configuration selected on the previous stop) and on switch, and they worked as expected.

PR #903, opened 2025-05-09
CRISTAL-515: Runtime sessions are not correctly closed with Blocknote

ClementEXWiki:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-512
https://jira.xwiki.org/browse/CRISTAL-515

# Changes

## Description

* (Potential) Fix for BlockNote crashes
* (Potential) Fix for multiplayer sessions not being closed properly

## Clarifications

* This is a single PR for two Jira issues as the two are interwined and too complex to separate
* The crashes happen in a non-deterministic way so it's impossible to tell if the bug is completely fixed or not, only time will tell
* The root cause for the bug is still unknown, there are mainly two things:
  * An `e.blocks is undefined` error in BlockNote, which is not supposed to be possible
  * Components mounted using React's `createRoot` sometimes end up being unmounted which is not supposed to be possible
* None of the two bugs above have been fixed ; but the code change in this PR seems to not trigger the actual root cause that makes BlockNote crash

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/core/reactivue/src/index.tsx
+++ b/core/reactivue/src/index.tsx
@@ -95,6 +95,9 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
       }
 
       // Render the element inside the container element
+      //
+      // NOTE: We put the root in a variable before assigning it to the component's state,
+      //       as to not access it through a Proxy, which could result in some weird issues
       const root = createRoot(this.$el);
       root.render(
         <ReactIndirectionLayer
--- a/core/reactivue/src/index.tsx
+++ b/core/reactivue/src/index.tsx
@@ -22,7 +22,7 @@
  */
 
 import { createRef, useEffect, useRef, useState } from "react";
-import { createRoot } from "react-dom/client";
+import { Root, createRoot } from "react-dom/client";
 import { createApp, defineComponent, h, toRaw } from "vue";
 import type { ReactElement, ReactNode } from "react";
 import type { App, SlotsType, VNode } from "vue";
@@ -81,6 +81,8 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
             options,
           ),
         ),
+        // The React root element
+        reactRoot: null,
       };
     },
 
@@ -93,12 +95,15 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
       }
 
       // Render the element inside the container element
-      createRoot(this.$el).render(
+      const root = createRoot(this.$el);
+      root.render(
         <ReactIndirectionLayer
           Component={Component}
           componentProps={this.$data.observableProps}
         />,
       );
+
+      this.$data.reactRoot = root;
 
       this.$watch(
         () => [{ ...this.$attrs }, { ...this.$slots }],
@@ -106,6 +111,16 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
           this.$data.observableProps.trigger();
         },
       );
+    },
+
+    beforeUnmount() {
+      const root = this.$data.reactRoot;
+
+      if (!root) {
+        throw new Error("Missing React root from adapter component's state");
+      }
+
+      root.unmount();
     },
 
     // The component will be rendered inside this element
@@ -119,6 +134,7 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
 type ReactAdapterComponentState<Props extends Record<string, unknown>> = {
   observableProps: Observable<Props>;
   options: ReactComponentAdapterOptions;
+  reactRoot: Root | null;
 };
 
 /**
--- a/core/reactivue/src/index.tsx
+++ b/core/reactivue/src/index.tsx
@@ -97,7 +97,10 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
       // Render the element inside the container element
       //
       // NOTE: We put the root in a variable before assigning it to the component's state,
-      //       as to not access it through a Proxy, which could result in some weird issues
+      //       as to not access it through a Proxy, which could result in some weird issues,
+      //       like some bugs deep in Y.js' dependency tree which were caused by some subtle
+      //       changes in behavior when accessing an object through a proxy instead of acecssing
+      //       the object directly.
       const root = createRoot(this.$el);
       root.render(
         <ReactIndirectionLayer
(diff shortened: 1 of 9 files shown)


core/reactivue/src/index.tsx
```diff
@@ -93,13 +95,16 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
       }
 
       // Render the element inside the container element
-      createRoot(this.$el).render(
+      const root = createRoot(this.$el);
```
manuelleduc, line 98: Since we are doing many indirections around `root`, feels better for future maintenance to explain at the different steps what we are doing and why.
Otherwise, changes are high that we create regressions during future refactorings.

editors/blocknote-headless/src/blocknote/uniast-to-bn.ts
```diff
@@ -1,3 +1,23 @@
+/*
```
manuelleduc, line 1: Should be in a separate [Misc] PR. (though thanks for nothing this!)

editors/blocknote-headless/src/react/BlockNoteView.tsx
```diff
@@ -91,15 +101,16 @@ type BlockNoteViewWrapperProps = {
...
- * @param blocks - the content to to load into the editor
+ * @param blocks - the content to load into the editor
  */
-async function parseAndLoadContent(
-  // TODO: MaybeUninit<EditorType>
-  editor: EditorType,
-  blocks: BlockType[],
-) {
+async function replaceContent(editor: EditorType, blocks: BlockType[]) {
+  // TODO: with time, see if this fix actually works fine
```
manuelleduc, line 109: Even if not fully sure, it's best to explain *why* you introduced this additional condition.

editors/blocknote-headless/src/vue/c-realtime-users.vue
```diff
@@ -40,6 +40,8 @@ watch(
       return;
     }
 
+    status.value = WebSocketStatus.Connecting;
```
manuelleduc, line 43: Same here, probably best to explain why the assignment of `status` is delayed.

editors/blocknote/src/vue/c-edit-blocknote.vue
```diff
@@ -81,6 +78,11 @@ const editorContent = shallowRef<UniAst | Error | null>(null);
 const editorInstance =
   useTemplateRef<InstanceType<typeof CBlockNoteView>>("editorInstance");
 
+/**
+ * Setup the editor and title input using the fetched page's content
+ *
+ * @param currentPage -
```
manuelleduc, line 84: missing comment

editors/blocknote/src/vue/c-edit-blocknote.vue
```diff
@@ -119,9 +125,40 @@ function view() {
   cristal?.getRouter().push(viewRouterParams);
 }
 
+/**
+ * Save a content into the current page document
+ *
+ * @param content -
```
manuelleduc, line 131: missing comment

editors/blocknote/src/vue/c-edit-blocknote.vue
```diff
@@ -119,9 +125,40 @@ function view() {
   cristal?.getRouter().push(viewRouterParams);
 }
 
+/**
+ * Save a content into the current page document
```
manuelleduc, line 129: Small detail, but we usually finish the tsdoc sentences with a dot.
Note to self: I should add a lint rule for this :)

editors/blocknote/src/vue/c-edit-blocknote.vue
```diff
@@ -198,10 +219,13 @@ watch(
         class="doc-title"
       />
     </template>
+
```
manuelleduc, line 222: Are the newlines needed in the template?

package.json
```diff
@@ -41,7 +41,7 @@
     "eslint-config-prettier": "10.1.2",
     "eslint-import-resolver-typescript": "4.3.4",
     "eslint-plugin-import": "2.31.0",
-    "eslint-plugin-prettier": "5.2.6",
+    "eslint-plugin-prettier": "5.3.1",
```
manuelleduc, line 44: seems unrelated to the current PR

package.json
```diff
@@ -64,6 +64,7 @@
     "ignoredBuiltDependencies": [
       "esbuild",
       "nx",
+      "unrs-resolver",
```
manuelleduc, line 67: seems unrelated to the current PR

package.json
manuelleduc: The changes in `pnpm-lock` are likely a consequence of this change and will disappear from the PR once fixed.

editors/blocknote/src/vue/c-edit-blocknote.vue
ClementEXWiki: I feel like they make the code more readable, instead of having a compact block of code where it's harder to see where blocks start and end

manuelleduc: thanks, if it's only for readability it's best in a separate PR

core/reactivue/src/index.tsx
```diff
@@ -93,13 +95,19 @@ function reactComponentAdapter<Props extends Record<string, unknown>>(
       }
 
       // Render the element inside the container element
-      createRoot(this.$el).render(
+      //
+      // NOTE: We put the root in a variable before assigning it to the component's state,
+      //       as to not access it through a Proxy, which could result in some weird issues
```
manuelleduc, line 100: Which kind of weird issues? Without more details, I don't think it's helping future readers.

editors/blocknote-headless/src/react/BlockNoteView.tsx
```diff
@@ -150,36 +164,64 @@ function BlockNoteViewWrapper({
...
+      );
+
+      provider?.disconnect();
+    };
+  }, []);
+
+  if (!ready) {
+    return (
+      <h3>
+        <em>Syncing changes with other realtime users...</em>
```
manuelleduc, line 221: Missing translation.

PR #927, opened 2025-05-23
CRISTAL-516: BlockNote is not WYSIWYG

ClementEXWiki:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-516

# Changes

## Description

* Make BlockNote look more like rendered pages

## Clarifications

A lot of things to say on this one!

* I tried different approaches, and ended up with the simplest route: using CSS to override style inside the BlockNote's HTML editor elements
* All style elements (font names, sizes, paddings, margins, etc.) use variables imported from the current Design System. This means BlockNote's inner style will now adapt to the selected DS.
* Bullet lists' markers cannot be made the same size as the rendered pages because they are unaligned HTML elements, so increasing their size leads to this:
image
* Lists of tasks are not consistent with the rendered pages as this would be too complex to do. But given they are not supported in XWiki's rendered pages, no one should theorically use them.
* The look'n'feel is not consistent with TipTap, as TipTap itself is not consistent with the rendered pages
* The current implementation is not ideal as we rely on some undocumented BlockNote selectors to change the style. Even though these _should_ remain the same, this is not guaranteed by BlockNote's maintainers
* The style of the rendered page is not always coherent, for instance level 4/5/6 headings all have the same size. This behavior has been replicated here as to closely match the rendered pages, but this it something that should be changed in the future.

# Screenshots & Video

Rendered page:

image

BlockNote:

image

BlockNote **before** this PR:

image

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/editors/blocknote-headless/src/vue/c-blocknote-view.vue
+++ b/editors/blocknote-headless/src/vue/c-blocknote-view.vue
@@ -328,11 +328,9 @@ const { t } = useI18n({
   }
 
   & [data-content-type="codeBlock"] {
-    /* background: var(--cr-color-neutral-100); */
     background: white;
     border-radius: var(--cr-border-radius-medium);
     font-family: var(--cr-font-mono);
-    /* padding: var(--cr-spacing-small); */
     color: var(--cr-base-text-color);
 
     & pre {
--- a/editors/blocknote-headless/src/vue/c-blocknote-view.vue
+++ b/editors/blocknote-headless/src/vue/c-blocknote-view.vue
@@ -308,6 +308,10 @@ const { t } = useI18n({
     font-size: var(--cr-font-size-medium);
   }
 
+  & [data-content-type="bulletListItem"] {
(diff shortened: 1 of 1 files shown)


editors/blocknote-headless/src/vue/c-blocknote-view.vue
```diff
@@ -273,4 +273,72 @@ const { t } = useI18n({
...
+    background-color: var(--cr-color-neutral-50);
+    color: var(--cr-color-neutral-600);
+    font-size: var(--cr-font-size-large);
+    border-inline-start: 2px solid var(--cr-color-neutral-200);
+    padding-inline-start: var(--cr-spacing-large);
+    margin: 0;
+  }
+
+  & [data-content-type="codeBlock"] {
+    /* background: var(--cr-color-neutral-100); */
```
manuelleduc, line 331: dead code

PR #934, opened 2025-05-28
Fix: don't overwrite others' work when editor starts empty

ClementEXWiki:
# Jira URL

N/A

# Changes

## Description

* Prevent overwriting other peoples' work when the editor starts empty for unknown reasons

## Clarifications

* This PR actually delegates the saving part to the component integrating the BlockNote editor component
* Realtime users and save status display are now delegated to the parent component as well

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


editors/blocknote/tsconfig.json
```diff
@@ -2,19 +2,12 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "dist",
-    "lib": [
-      "dom", "ESNext"
-    ],
+    "lib": ["dom", "ESNext"],
```
manuelleduc, line 5: Are those changes related to the current PR?

pnpm-lock.yaml
manuelleduc, line 1: I don't understand why `pnpm-lock` is changed.

editors/blocknote/src/vue/c-save-status.vue
```diff
@@ -17,58 +17,52 @@ License along with this software; if not, write to the Free
 Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 -->
+<script lang="ts">
```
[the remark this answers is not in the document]
ClementEXWiki: You can't export anything from a `setup` script in a Vue component, so you need a dedicated block script for exporting.

editors/blocknote/tsconfig.json
manuelleduc: ok, it should be in a separate PR then

editors/blocknote/src/vue/c-save-status.vue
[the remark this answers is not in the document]
manuelleduc: Wouldn't it be cleaner in a separate ts file?

pnpm-lock.yaml
ClementEXWiki: Because there was a transitive dependency version conflict between TipTap and BlockNote and some Y.js sub-package

editors/blocknote/tsconfig.json
manuelleduc: ok I checked further and IntellijJ and VSCode are not formatting this file in the same way.
That's another argument to deal with this separately.

pnpm-lock.yaml
manuelleduc, line 1: looks much better now :)

PR #1200, opened 2025-09-08
CRISTAL-617: Show a message before leaving with unsaved content in BlockNote

ClementEXWiki:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-617

# Changes

## Description

* Show a confirmation dialog when quitting BlockNote with unsaved changes
 
## Clarifications

* Only applies to the full-blown BlockNote editor package, not to the `blocknote-headless` part given that each integration may desire to customize this behavior

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/editors/blocknote/langs/translation-en.json
+++ b/editors/blocknote/langs/translation-en.json
@@ -1,3 +1,4 @@
 {
-  "blocknote.editor.save.error": "Failed to save the page."
+  "blocknote.editor.save.error": "Failed to save the page.",
+  "blocknote.editor.save.unsavedChanges": "You have unsaved changes. Do you really want to quit?"
 }
--- a/editors/blocknote/src/vue/c-edit-blocknote.vue
+++ b/editors/blocknote/src/vue/c-edit-blocknote.vue
@@ -37,7 +37,7 @@ import {
 } from "@xwiki/cristal-uniast-markdown";
 import { createConverterContext } from "@xwiki/cristal-uniast-utils";
 import { debounce } from "lodash-es";
-import { inject, ref, shallowRef, useTemplateRef, watch } from "vue";
+import { inject, onMounted, ref, shallowRef, useTemplateRef, watch } from "vue";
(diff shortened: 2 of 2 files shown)


manuelleduc:
Unless I missed something, I tried the following steps:
1. block the save url from the network tab of my browser debugger
2. edit and input some content on a page
3. wait for an error message for the save fail
4. click on any UI element that can make me lost the current editor content (breadcrumb, navigation tree, closing the tab...)

I never get a warning message for the potential data loss.

ClementEXWiki:
Good catch, should be fixed now.

ClementEXWiki:
> > Can't reproduce on my side, when closing the window, I do get a warning message.
> 
> What method did you use to reproduce? In my case I use the XWiki backend and blocked the save URL with the debugger of Electron.

I used Chrome and put the website offline using the network devtools. Then I edited the page, and tried the following:

* Closing the page with the "X" tab icon
* Closing the page with Ctrl+W
* Closing the window with the "X" window button
* Closing the window with Alt+F4
* Navigating to another page using the left side navigation tree

All of which show warnings. 

I tested quickly with Electron, and clicking on the navigation tree does bring the warning popup as well.

manuelleduc:
> Right, it doesn't show a message on Electron when closing the window, only on navigation. Seems related to [electron/electron#2301](https://github.com/electron/electron/issues/2301)

I found https://jira.xwiki.org/browse/CRISTAL-429 I had the same issue when working on it for Tiptap apparently.

> Does that mean we put some Electron-specific logic inside that component? I'm not fond of it but I don't really have another idea, given Electron doesn't seem to follow web browsers' standard behavior.

I'm going to merge as is to not block the release but when working on CRISTAL-429, we'll need to find an architecture to decouple the close action from the platform (electron/browser) it's running on. So yes we don't want to put hardcoded  electron specific cases.

PR #1253, opened 2025-09-29
CRISTAL-49: add alert macros

ClementEXWiki:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-49

# Changes

## Description

* Add alert macros (information, warning, error) in BlockNote

## Clarifications

* As macros are not supported in the HTML renderer yet, they are not visible in view mode, only in BlockNote
* As macros don't support inner content yet, the alerts cannot be edited in a WYSIWYG manner
* As macros don't support nested content, only plain text is supported inside the alerts
* As there is no edition modal for Cristal yet, the alerts cannot be edited **_at all_** in BlockNote for now

# Screenshots & Video

image

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


editors/blocknote-react/src/blocknote/macros/AlertMacros.tsx
```diff
@@ -0,0 +1,71 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+import { createMacro } from "../utils";
+import type { CSSProperties } from "react";
+
+function createAlertMacro(
```
manuelleduc, line 24: To align with XWiki's vocabulary, I suggest using the keyword "message" instead of "alert". Also, I find it confusing to have a success alert macro :)

manuelleduc (changes requested):
This PR is missing many essential features:
- the content is not editable (and possibly other parameters)
- I don't understand how the macro will be displayed in view mode
- the macros must be declared outside of the blocknote package, because they are not part of any specific editor (or any displayer), and must be provided through an extension mechanism
- we must also make sure to align with the underlying backend (e.g., the current "Info" macro is not parsable by the xwiki backend). Possibly by providing configurable mapping for each backend

editors/blocknote-react/src/index.tsx
```diff
@@ -78,6 +83,9 @@ function mountBlockNote(
 const DEFAULT_MACROS = {
   XWikiMacroHTMLBlockMacro,
   XWikiMacroInlineHTMLMacro,
+  InfoMacro,
```
manuelleduc, line 86: This list must be loaded by an extension mechanism

PR #1262, opened 2025-09-30
CRISTAL-626: Blocknote does not support the absence of link suggest for the current backend

ClementEXWiki:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-626

# Changes

## Description

* Display an information message in the search results when search is not supported for the current provider (the message is hidden when inputting an URL)
* Always accept entering raw entity references in the input field

## Clarifications

N/A

# Screenshots & Video

image

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/editors/blocknote-react/src/components/SearchBox.tsx
+++ b/editors/blocknote-react/src/components/SearchBox.tsx
@@ -23,7 +23,7 @@ import { debounce } from "lodash-es";
 import { useCallback, useEffect, useState } from "react";
 import { RiLink } from "react-icons/ri";
 import type { LinkEditionContext, LinkSuggestion } from "../misc/linkSuggest";
-import type { ReactElement } from "react";
+import type { KeyboardEvent, ReactElement } from "react";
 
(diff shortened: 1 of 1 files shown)


manuelleduc (commented):
Thanks @ClementEXWiki ! The code looks good!

I just found a small cornercase issue. When inserting an internal reference (e.g., `Sandbox.WebHome` on an XWiki backend), the link is serialized as an external link (i.e., `[blabla](Sandbox.WebHome)` instead of `[[blabla|Sandbox.WebHome]]`).

I cheated by editing the body of `DefaultLinkSuggestServiceProvider#get` to always return `undefined` during my tests.
I'm not expecting the fix to be too impactful, but let me know if I'm too optimistic.

manuelleduc:
@ClementEXWiki I think you should be able to close https://jira.xwiki.org/browse/CRISTAL-656 as "fixed by" CRISTAL-626 once this RP is merged.

PR #1635, opened 2026-04-09
CRISTAL-464: User should not be able to create pages when not logged in

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/CRISTAL-464

# Changes

## Description

This PR adds a check in the page creation menu to make sure the user has edit rights on the parent page. If not, display an appropriate message.
This also includes a UI test.

## Clarifications

We cannot simply hide the "New Page" button, as suggested in the issue, because it's not a "quick add" feature but Cristal's main page creation feature, allowing to select another parent. Also, for XWiki specifically, Guest could be allowed page creation for some spaces.
Ideally, the available spaces in the page creation menu should depend on the user's rights, but this is another issue (https://jira.xwiki.org/browse/CRISTAL-689) that is a lot harder to fix.

# Screenshots & Video

N/A

# Executed Tests

Tested manually and with the newly added test.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


pjeanjean:
Note: I had to change the mock server to add edit rights on the main page (for the existing page creation test). This had an impact on the screenshots. I contemplated adding a way to chose whether or not the main page should display with edit rights, but in the end I decided to update the screenshots instead since it allows us to test more of the interface.

skin/src/vue/c-page-creation-menu.vue
```diff
@@ -79,14 +80,30 @@ async function createPage() {
     newDocumentName,
     location.value!,
   );
+  const parentDocumentReference = referenceHandler.createDocumentReference(
+    location.value!.names[location.value!.names.length - 1],
```
manuelleduc: It feels weird to have to parse the location value like this.
Don't we have a more generic API that does the job based on each backend rules?

pjeanjean: We do not so far, but it would be useful. Maybe a `getParentDocumentReference` in the referenceHandler?

manuelleduc: I think so yes

PR #1686, opened 2026-05-27
XWIKI-24231: Expose the insert macro action

ClementEXWiki:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-24231

Sister PR in xwiki-platform: https://github.com/xwiki/xwiki-platform/pull/5539

# Changes

## Description

* Add a macro insertion mechanism in the toolbar in toolbar and in the quick actions dropdown

## Clarifications

* Just like for link edition, Cristal only has a placeholder for now, actual implementation depends on the caller

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  *

### DIFF
--- a/editors/blocknote/src/vue/c-edit-blocknote.vue
+++ b/editors/blocknote/src/vue/c-edit-blocknote.vue
@@ -136,6 +136,11 @@ const contextForMacros: ContextForMacros = {
   openParamsEditor(/*macro, params, update*/) {
     alert("TODO: params editor for macros in Cristal");
   },
+
+  openInsertionEditor(prefill) {
+    console.debug({ prefill });
+    alert("TODO: insertion editor for macros in Cristal");
+  },
 };
 
 /**
@@ -332,7 +337,7 @@ onBeforeRouteLeave(() => {
                 ref="editorInstance"
                 :editor-props
                 :editor-content
-                :container
+                :deps-container="container"
                 :collaboration
                 :macros="{
                   ctx: contextForMacros,


editors/blocknote/src/vue/c-edit-blocknote.vue
```diff
@@ -136,6 +125,11 @@ const contextForMacros: ContextForMacros = {
   openParamsEditor(/*macro, params, update*/) {
     alert("TODO: params editor for macros in Cristal");
   },
+
+  openInsertionEditor(prefill) {
```
manuelleduc, line 140: I don't understand what this method is needed for:
1. it lack documentation
2. if it is made to be extended, I find it surprising that it does not impact the package public API
3. I don't think we should use alert, when is this code called? We need to find a better way to communicate to the user that this feature is not supported
4. We shouldn't mention "Cristal" in xwiki-platform codebase (we have leftover from the move but the goal is to remove them, not to add more). I suggest mentioning something about the need for the method to be extended by other packages reusing this code
5. We can probably make the debug log clearer by adding an explanatory text in addition to the parameter?

ClementEXWiki: I don't really get what you're saying - this function is defined in `xwiki-platform` and has nothing to do with Cristal. It is required that the usage of BlockNote editor defines two functions for macros: one for opening the parameters edition modal, and one for the insertion modal. We already have the other one.

I don't see any mention of Cristal in the `xwiki-platform` PR I created.

PR #1715, opened 2026-07-16
CRISTALNC-36: New page modal parent location field

manuelleduc:
# Jira URL

https://jira.xwiki.org/browse/CRISTALNC-36

# Changes

## Description

* The breadcrumb displaying the parent location in the new page (and move page) dialog was made of real links: clicking a segment navigated to that location while leaving the dialog open. The urls are now stripped from the breadcrumb items so the segments are not rendered as links (`url` is already optional in `BreadcrumbProps`). The location display keeps using the abstract `x-text-field` design system component (readonly, breadcrumb in the default slot), as before.
* Choosing a location in the "Change Page Location" dialog now only stages the choice; it is applied to the field once the user confirms with the Select button (and discarded if the dialog is closed without confirming), instead of being applied immediately on the tree node click.
* The new page dialog content used a hard `min-width: 600px`, which overflowed with a horizontal scrollbar in a dialog narrower than 600px (e.g. the fixed-width Nextcloud dialog). It now targets 600px but caps at `max-width: 100%`, so it keeps the intended width in the auto-sizing Vuetify dialog while fitting a narrower dialog without overflow.
* The new page dialog now opens with the parent location set to the current page, refreshed on each opening. Previously the location was computed once when the menu mounted (so it went stale after navigating), and was set to the *space* of the current document — which for path-based backends (e.g. Nextcloud, where the document name is a plain path segment) designated the parent of the current page (or even the root for top-level pages) instead of the page itself. `NavigationTreeSelect` now honors a location provided by its caller instead of always overwriting it with the current page space (`MovePage` keeps the previous default, which is correct for moving).

## Clarifications

* This PR addresses the behavioral problems of CRISTALNC-36 (breadcrumb navigation with the dialog open, horizontal overflow, plus the selection and default-location improvements). The purely cosmetic points of the report (input affordances on the readonly field, cropping of long locations, hover background overlapping the label in the Nextcloud design system) come from how each design system implementation of `x-text-field` renders a readonly field and places its default slot (Vuetify: field content, Shoelace: prefix, Nextcloud: icon slot); addressing them belongs in the design system implementations, not in this component.
* No design-system API change.
* The Nextcloud app will pick the fixes up with its next dependency bump.

# Screenshots & Video

The issue contains a video of the previous behavior. After the change, the location field looks as before (readonly design system text field with the breadcrumb inside), but its segments are not clickable, the location shown is the current page and only changes when Select is pressed, and the dialog no longer shows a horizontal scrollbar in the Nextcloud design system.

# Executed Tests

* `pnpm --filter ./core/navigation-tree/navigation-tree-ui test` (6 unit tests for `NavigationTreeSelect`: readonly field display, breadcrumb segments without urls, caller-provided location honored, location committed only on Select, last staged choice committed, nothing committed when Select is pressed without a choice), plus `lint`, `typecheck`, `build` and `api-extractor:local` on the same package.
* `pnpm --filter ./skin test` — new `c-page-creation-menu` tests checking the dialog opens with the current page as parent location for both path-based and XWiki-style (`WebHome`) backends, plus the pre-existing skin tests.
* Manually verified with `pnpm run dev:cristal-web` against the XWiki demo backend (Playwright-driven): opened the new page dialog, changed the location through "Change Page Location", checked that the breadcrumb updates, has no links, and that clicking a segment neither navigates nor closes the dialog. Confirmed the dialog content stays 600px wide in the Vuetify design system, and that after navigating in-app from the home page to another page, reopening the dialog shows the newly current page as parent location.
* The horizontal-scrollbar fix was reproduced and validated against the actual Nextcloud design system (`NcDialog`, `@nextcloud/vue`) in a standalone harness: the `min-width: 600px` content overflowed the ~388px fixed dialog before the change and fits without overflow after.

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches: none

### DIFF
--- a/core/navigation-tree/navigation-tree-ui/package.json
+++ b/core/navigation-tree/navigation-tree-ui/package.json
@@ -46,11 +46,16 @@
   },
   "devDependencies": {
     "@types/async-lock": "catalog:",
+    "@vue/test-utils": "catalog:",
     "@xwiki/cristal-dev-config": "workspace:*",
+    "@xwiki/cristal-dev-test-utils": "workspace:*",
     "@xwiki/platform-tool-apiextractorconfig": "catalog:",
     "@xwiki/platform-tool-tsconfig": "catalog:",
+    "inversify": "catalog:",
     "typescript": "catalog:",
     "vite": "catalog:",
+    "vitest": "catalog:",
+    "vitest-mock-extended": "catalog:",
     "vue": "catalog:",
     "vue-tsc": "catalog:"
   },
--- a/core/navigation-tree/navigation-tree-ui/src/vue/NavigationTreeSelect.vue
+++ b/core/navigation-tree/navigation-tree-ui/src/vue/NavigationTreeSelect.vue
@@ -21,7 +21,7 @@
 import NavigationTree from "./NavigationTree.vue";
 import messages from "../translations";
 import { SpaceReference } from "@xwiki/platform-model-api";
-import { inject, onBeforeMount, ref, watch } from "vue";
+import { computed, inject, onBeforeMount, ref, useId, watch } from "vue";
 import { useI18n } from "vue-i18n";
 import type { CristalApp } from "@xwiki/platform-api";
 import type { DocumentService } from "@xwiki/platform-document-api";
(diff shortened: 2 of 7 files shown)


manuelleduc:
Merged as 594d3c4ef8fe67f59b5d81c2ccfc6ae89ab5c534.

